Issue #43, opened 2014-03-15
Questions about html escape


dstrelau:
Hi folks. I'm getting unexpected behavior when trying to send formatted links.

I'm attempting to send a response from hubot with a link label, like so:

```
msg.send "<https://github.com/link/to/a/PR|myrepo #42> fix some broken"
```

What I'd expect to see in Slack is this, with `{{ … }}` representing hyperlinked text with the URL I passed.

```
{{myrepo #42}} fix some broken
```

Instead, I see this, with literal brackets and the link in the wrong place:

```
<{{myrepo}} #42> fix some broken
```

Looking around, it's pretty evident that this is due to [the `escapeHTML` call](https://github.com/tinyspeck/hubot-slack/blob/master/src/slack.coffee#L28) that occurs before the message is sent. In fact, [removing that call](https://github.com/dstrelau/hubot-slack/commit/3ff476e59ca01a6634a7160d740ad837f2f98528) gives me the result I'd expect, with the properly parsed link.

I'm trying to understand how https://api.slack.com/docs/formatting works in the context of hubot, but I'm a bit confused. What is the best way to allow for what I want to do without screwing up everyone else's hubot? Should I create `msg.sendRaw`? Is it actually safe to remove `escapeHTML` since slack handles most of it server-side?

dstrelau:
@paulhammond Could you provide an update here? I'd be happy to work on the adapter code here if the API fixes you mentioned have taken place.

lukerollans:
Also experiencing this issue! When our hubot sends pull request related notices to Slack, the message should appear as `repo#pr_number`, eg `my-repo#35`. This means that the msg is sent to Slack like ...

`https://github.com/organization/repo/pull/25|repo#25', however the pipe character is being url encoded so the entire message appears broken.

Watching this issue closely!

technicalpickles:
I was able to reproduce this pretty easily. I'm not sure why it's being escaped yet, as there doesn't to seem to be any obvious escaping code in the adapter. It might be in the node-slack, or maybe it's a parameter to the API that needs to be specified to parse it as mrkdwn.

I found a workaround though. It's not obvious, but you can use slack attachments to do this:

```coffeescript
robot.emit 'slack.attachment', {text: "<https://github.com/link/to/a/PR|myrepo #42> fix some broken">}
```

technicalpickles:
[...]

PR #89, opened 2014-10-16
don't escape messages

wyefei:
We made the quick change to fix the issue #43 and since there's no test, I guess this is o.k


tedsteen:
@paulhammond where is that thread? This is still a problem and it's hard to find a workaround.

PR #133, opened 2014-12-18
Create new Message types for slack messages

kballard:
Regular text messages now use a TextMessage subclass called
SlackTextMessage, which includes both the parsed text (as `text`) and
the unparsed text (as `rawText`).

Other messages that used to be ignored are now received as either
SlackRawMessage or SlackBotMessage. These do not descend from
TextMessage so they won't be seen by normal `hear` or `respond`
listeners.

A few new Listener classes were created, SlackRawListener and
SlackBotListener, to make it slightly easier to write code tailored for
these new message types.

This PR is based on #131.

### DIFF
--- a/index.coffee
+++ b/index.coffee
@@ -1,0 +1,15 @@
+SlackBot = require './src/slack'
+{SlackTextMessage, SlackRawMessage, SlackBotMessage} = require './src/message'
+{SlackRawListener, SlackBotListener} = require './src/listener'
+
+module.exports = exports = {
+  SlackBot
+  SlackTextMessage
+  SlackRawMessage
+  SlackBotMessage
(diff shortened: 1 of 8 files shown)


paulhammond:
Sorry, I was hoping to work on Hubot-slack over the holidays but was unexpectedly offline and couldn't. I just merged this - thank you for the code, and for responding to all the feedback!

kballard:
@paulhammond Thanks. Can I encourage you to push a new version so I can start depending on this stuff in my bot?

technicalpickles:
I had someone in #hubot ask about using this, which makes me think it needs some documentation cc https://github.com/slackhq/hubot-slack/issues/245

PR #137, opened 2015-01-06
Use require.main.require 'hubot' instead of require 'hubot'

kballard:
Because we have hubot in our devDependencies, if these dependencies are 
installed, `require 'hubot'` will pick up the local version of hubot instead
of the one that the robot is running from. This means that our emitted
`TextMessage`s won't trigger any `TextListener`s installed by scripts (since
that uses `instanceof`).

As a workaround, use `require.main.require 'hubot'` instead, which requires
hubot from the context of the main module of the program. This should produce
the same module that the robot is running from (barring unusual edge cases).
If that fails, fall back to `require 'hubot'`, to handle cases such as running
the tests.

### DIFF
--- a/src/hubot.coffee
+++ b/src/hubot.coffee
@@ -1,0 +1,18 @@
+# Because we have hubot in our devDependencies, if these dependencies are
+# installed, `require hubot` picks up the local version of hubot instead of the
+# one the robot is running from. This means that our emitted `TextMessage`s
+# won't trigger any `TextListener`s installed by scripts as that uses
+# `instanceof` (which doesn't handle duplicate definitions).
+#
+# As a workaround, we can use `require.main.require 'hubot'` to require 'hubot'
+# from the perspective of the "main" module. This will produce the correct
+# version of hubot under normal circumstances. This should only fail if hubot
+# is being used as a component of a larger application (even though it's not
+# designed to work that way).
+
+try
+  module.exports = require.main.require 'hubot'
+catch
+  # If that failed, we'll assume it's because it couldn't find the hubot module.
+  # In that event, let's just fall back to a normal require.
+  module.exports = require 'hubot'
--- a/src/listener.coffee
+++ b/src/listener.coffee
@@ -1,4 +1,4 @@
-{Listener} = require 'hubot'
+{Listener} = require './hubot'
 {SlackRawMessage, SlackBotMessage} = require './message'
 
 class SlackRawListener extends Listener
--- a/src/message.coffee
+++ b/src/message.coffee
@@ -1,4 +1,4 @@
-{Message, TextMessage} = require 'hubot'
+{Message, TextMessage} = require './hubot'
 
 # Hubot only started exporting Message in 2.11.0. Previous version do not export
 # this class. In order to remain compatible with older versions, we can pull the
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -1,4 +1,4 @@
-{Robot, Adapter, EnterMessage, LeaveMessage, TopicMessage} = require 'hubot'
+{Robot, Adapter, EnterMessage, LeaveMessage, TopicMessage} = require './hubot'
 {SlackTextMessage, SlackRawMessage, SlackBotMessage} = require './message'
 {SlackRawListener, SlackBotListener} = require './listener'
 


kballard:
Related: I wish npm supported installing `devDependencies` somewhere other than `node_modules` so your `devDependencies` couldn't interfere with the behavior of `require` in the main application.

evansolomon:
It seems like hubot should just be a normal dependency since it's used in normal code

kballard:
If it's a normal dependency then that would break this even more. `hubot-slack` does not want to include `hubot` as a dependency under normal operations, because it's meant to be a dependency *of* `hubot`.

Ideally, `hubot` would pass its own module as a second argument to `use`, thus providing an explicit reference to the module for adapters to use (and the adapter can avoid loading up its code until `use()` is called). But it doesn't do that.

kballard:
To clarify what's going on here:

Node handles `require 'foo'` by looking in `./node_modules/` for `foo`, and walking up the parent tree if it can't find `'foo'`. If it can't find `node_modules/foo/` in any parent directory, it throws an error.

This search is always done relative to the current module. This means that two separate modules loaded into the same app, that both have a dependency on a module `'bar'` and thus both contain `node_modules/bar`, will actually load distinct copies of the `'bar'` module. Not only is all the code loaded twice, but since all types are defined twice, `instanceof` won't work properly when one module calls it on a value produced by the other module. This is similar to why `instanceof Array` doesn't work in various circumstances and thus why `util.isArray()` exists.

As long as you're ok with duplicate copies of the code being loaded, and nobody's using `instanceof`, this usually isn't problematic. And in fact it's often very beneficial for these duplicated dependencies to be loaded separately; among other reasons, it allows for loading different versions of the same module.

However, in the case of `hubot-slack`, this module provides a component that `hubot` uses. Having `hubot` be a dependency of `hubot-slack` doesn't make sense, because then you end up with the adapter using the wrong copy of the module. Unfortunately, we need to have `hubot` be a `devDependency` in order for the tests to work. And since npm installs `devDependency`s into the same `node_modules` folder shared by regular dependencies, this means that installing the `devDependency`s actually affects the behavior of the "real" module.

It's also worth noting here that if `hubot-slack` is installed via `npm link`, then if `devDependency`s haven't been installed, `require 'hubot'` will actually fail because `node_modules/hubot` will not exist in any parent folder (except in pathological setups). This is because the module is resolved from the "real" path to `hubot-slack`, as opposed to the location of the symlink created by `npm link`.

kballard:
There is a separate approach we could take instead of `require.main.require`, which is to have `index.coffee` use `module.parent` to get a reference to the module that imported it. This would be `hubot/src/robot`. However, doing this would require having an implicit dependency on the implementation details of the `hubot` module, because we'd have to know to use `module.parent.parent` to get the actual `hubot` module, and it would also cause problems if anyone else ever requires `hubot-slack` besides `hubot` itself (such as the tests).

As an alternative, we could also try calling `module.parent.require 'hubot'`, and continuing up the parent chain if that fails, in order to find the closest including module that can locate a copy of `hubot` outside of `hubot-slack`. Under normal circumstances `module.parent.require 'hubot'` would be sufficient, because `hubot/src/robot` would be able to find `hubot` in `hubot/src/../../../node_modules/hubot`. This would of course fail with `npm test` the same way `require.main.require 'hubot'` does and require the fallback to `require 'hubot'`. However, I felt that `require.main.require 'hubot'` was conceptually simpler.

evansolomon:
> If it's a normal dependency then that would break this even more

Can you expand on this?  From a quick look at the code it's not clear to me why it would break.  It looks like the only thing the hubot module is used for is to extend classes (which don't have any state).

kballard:
If it's a normal dependency then `require 'hubot'` will *always* find and load the local `hubot` dependency instead of finding the already-loaded `hubot` module that is loading the adapter.

Having two copies of `hubot` loaded is a problem precisely because `hubot` uses `instanceof` when testing if a given emitted message is a `TextMessage`, and `instanceof` is one of those things that breaks when you load multiple copies of a module. When `hubot-slack` loads a copy of `hubot` that's distinct from the one that produced the `Robot` instance that represents the bot, none of the `TextListener`s will work on any of the messages emitted by `hubot-slack`, meaning `robot.hear` and `robot.respond` will break.

`hubot-slack` only works in the general case right now because its `devDependency`s are not installed when it's installed as a dependency. But if you take a checkout of `hubot-slack` that has the `devDependency`s installed, and use `npm link` to link it globally, then go to your Hubot checkout and use `npm link` to make it use your linked `hubot-slack` instead of a "fresh" copy, your scripts will probably all break.

evansolomon:
> hubot uses instanceof when testing if a given emitted message is a TextMessage

Aha! This is the part I needed, thanks.

evansolomon:
The proposed solution is clever.  I wonder if maybe it's a little *too* clever and there are plausible scenarios where `require.main` isn't going to be what we think it is.  I could imagine that sort of thing leading to this is failing in a confusing way.

The other confusing form of failure could be when you end up with devDependencies installed in a non-dev environment.  There are a whole bunch of ways this could happen (e.g. ssh in, do some manual tests, etc), in which case you'll end up with this wrong hubot scenario and no clear feedback about how to fix it.  Semantically, it seems like what we're talking about is actually close to npm's peerDependencies.

@kballard thanks for your detailed report + explanation.  Would be curious to hear others' thoughts.  Personally I feel like I need to think a bit more about it to have a well-informed opinion.

kballard:
> there are plausible scenarios where `require.main` isn't going to be what we think it is

I can't think of any good reason for that to be true. The only two cases I can think of where this might happen are

1. When running the tests. The fallback to `require 'hubot'` exists for that case, and works fine.
2. When the `hubot` is loaded as part of a larger application. I don't think this is particularly valid though, as Hubot is designed to run as the application itself (notably, it calls `process.exit` and it binds an `express` instance to `$PORT`, and there may be other cases as well).

> The other confusing form of failure could be when you end up with devDependencies installed in a non-dev environment. There are a whole bunch of ways this could happen [...], in which case you'll end up with this wrong hubot scenario and no clear feedback about how to fix it. 

That's precisely what this PR is intended to fix. Having it installed in a non-dev environment is the same as having it installed in a dev environment that's being used to run a bot (e.g. via `npm link`). The PR fixes this case, making it so accidentally installing `devDependency`s in a non-dev environment doesn't cause the mysterious bot breakage.

> Semantically, it seems like what we're talking about is actually close to npm's peerDependencies.

npm's Peer Dependencies are something I haven't seen before. Reading up it, it does sound very similar. However, it seems that Peer Dependencies really only solves the case of needing to require that the sibling package matches a given version requirement. It doesn't solve the `devDependency`s issue that we're having, which is caused by `hubot-slack` needing to actually call `require('hubot')` and needing this to work within the context of `npm test`. Plausibly, we could hack up the test runner to create a stub Hubot package and expose a hook for it to hand this stub package to `hubot-slack` instead of calling `require 'hubot'`.

Really, the correct solution is for NPM to put `devDependency`s somewhere besides `node_modules`, and to have Node.js support alternate load paths. But I don't see that happening any time soon.

evansolomon:
Oops, good point re: the dev-in-prod issue.  I got mixed up on the order of the try/catch.

kballard:
I've submitted #138 as an alternative approach. That version is more complicated, but should be more robust to edge cases.

kballard:
A final approach, that I abandoned in favor of #138, is to remove `hubot` from the `devDependency` list altogether and write a stub module that provides enough functionality to stand in for hubot, and then muck with the `require.extensions` list in order to rewrite the module source before it's evaluated to add the stub module to the `module.paths` list. I abandoned it in favor of #138 as I felt that was a simpler approach.

And yet another approach would be to use the same stub module but stuff its path into the `globals` object and then have the library check `globals` for the stub path before calling `require 'hubot'`. But using global variables made me too uncomfortable.

johnagan:
I think `require.main.require 'hubot'` is the answer here, excluding the relative module reference. I'm adding this logic to our v3 branch for review. 

Thanks for the help @kballard

aoberoi:
adding a reference to a possibly related issue: #283

PR #138, opened 2015-01-06
Avoid our devDependencies when requiring 'hubot'

kballard:
Because we have hubot in our devDependencies, if these dependencies are 
installed, `require 'hubot'` will pick up the local version of hubot instead
of the one that the robot is running from. This means that our emitted
`TextMessage`s won't trigger any `TextListener`s installed by scripts (since
that uses `instanceof`).

Implement a workaround that avoids loading 'hubot' from our local node_modules
folder if possible. It first tries finding 'hubot' without our node_modules,
and if that fails, it walks up the module parent tree until it finds a module
outside our directory tree and tries again from that spot. This means it works
correctly even if hubot-slack is installed with `npm link`. It falls back to a
normal `require 'hubot'` if all else fails, which allows the test suite to
work.

This is an alternative to PR #137.

### DIFF
--- a/src/hubot.coffee
+++ b/src/hubot.coffee
@@ -1,0 +1,49 @@
+# Because we have hubot in our devDependencies, if these dependencies are
+# installed, `require hubot` picks up the local version of hubot instead of the
+# one the robot is running from. This means that our emitted `TextMessage`s
+# won't trigger any `TextListener`s installed by scripts as that uses
+# `instanceof` (which doesn't handle duplicate definitions).
+#
+# As a workaround, all requires of hubot are funneled through this module, and
+# here we can mess with our module.paths to ensure our devDependency doesn't
+# affect the loading of hubot.
+
+Path = require 'path'
+
+oldPaths = module.paths.splice(0) # make a copy
+modRoot = Path.dirname __dirname
+skipPath = Path.join modRoot, 'node_modules'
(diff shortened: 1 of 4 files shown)


technicalpickles:
What you are looking for is peerDependency. That will use the hubot dependency that is installed alongside hubot-slack in the project's node_modules

kballard:
I think we need to adopt `peerDependency`, but that doesn't actually solve the problem. `peerDependency` is purely a npm thing, it doesn't affect node.js's resolve rules, and we still need the `devDependency` in order to run `npm test`.

johnagan:
:wave: @kballard Thank you for putting in so much effort on fixing this here and #137. @DEGoodmanWilson and I are still catching-up on these PRs, so excuse me if I missed some of the discussions, but I'm curious your thoughts on using"dyl" : "file:../dyl"

DEGoodmanWilson:
TBH, I'm not in love with this solution or #137 . But, believe you me, I feel your pain, acutely, having run into precisely these issues myself trying to build and test `hubot-slack` using `npm link`.

I find it curious that other Hubot adapters don't list Hubot as a devDependency, and am trying to figure out how that even works.

So, I'm meditating on this PR and #137 for a bit…please know we're paying attention and thinking on these things, and thank you for all the thought you've put into this issue as well.

Issue #136, opened 2014-12-31
robot.messageRoom(chan, msg) doesn't work if chan starts with "#"


yozlet:
... which wouldn't be a problem if it hadn't worked in v2, but it did, and it took a while to work out why one of my scripts wasn't working any more.

trodrigues:
Any idea when we'll see the slack-client version updated here to fix this?

paulhammond:
Pretty soon - we'll need to bump the version when we merge #148, so I'll likely do it at the same time

technicalpickles:
https://github.com/slackhq/hubot-slack/pull/148 was merged, so this can be closed.

PR #139, opened 2015-01-09
replace "#" from channel name prefix

shokai:
This is an alternative implementation for #136

Some hubot-scripts using `robot.send("#room_name", msg)`.
It was ok at hubot-slack v2, but not works with current version (v3).

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -191,7 +191,7 @@ class SlackBot extends Adapter
     text
 
   send: (envelope, messages...) ->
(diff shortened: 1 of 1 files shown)


paulhammond:
Thanks for your code, and sorry to take so long to review it! Since you submitted this pull request the bug was fixed in node-slack-client (slackhq/node-slack-client#9) so this change is no longer needed! Yay!

shokai:
:beers:

PR #140, opened 2015-01-09
direct reply when using the `reply`

plukevdh:
hellocomputer

Instead of just replying with a username, why not make this more useful for quieting slack down a bit. 

Seems like the reply method is more of a holdover from chat clients w/o the DM functionality. `send` and `reply` are almost identical otherwise. Is this useful to anyone else? Open to other suggestions on how to make this happen.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -232,11 +232,10 @@ class SlackBot extends Adapter
         channel.send m for m in submessages
 
   reply: (envelope, messages...) ->
-    @robot.logger.debug "Sending reply"
+    @robot.logger.debug "Sending direct reply"
 
-    for msg in messages
-      # TODO: Don't prefix username if replying in DM
-      @send envelope, "#{envelope.user.name}: #{msg}"
+    user = @client.getDMByName envelope.user.name
+    user.send msg for msg in messages
 
   topic: (envelope, strings...) ->
     channel = @client.getChannelGroupOrDMByName envelope.room


evansolomon:
Is the idea here that if I do something that ends up triggering `msg.reply` somewhere I will get a DM instead of an `@evan` wherever the original message was sent?

I think that seems a bit weird since I can already DM the bot if I want it to reply via DM.

plukevdh:
That is the desired behavior in this PR. The idea is to mimic some of what the built-in Slackbot does by replying in a user-only visible way, giving us the ability to hide actions needing to be taken by the requesting user. OAuth is a good example of this kind of action. Checkout the google hangout integration and how it works. This kind of behavior would allow us to mimic some of that in Hubot-backed robots.

paulhammond:
I don't think we're going to merge this as is. In almost all cases people expect a reply to go back to the same channel as the original message. In the cases where you'd like to reply directly to the user that's still possible (although it will need some work in the script).

And, there's nothing stopping you applying this patch to your hubot yourself if you'd like it work this way in your team (yay for open source!)

If I'm missing a reason for this to be the default case please let me know!

pklapperich:
The official hubot documentation states:
"With it, you can send a message back to the room the res came from, emote a message to a room (If the given adapter supports it), or reply to the person that sent the message."

This indicates that reply should reply to the PERSON, not the channel. The script I'm working on will respond with sensitive information that other participants in the channel shouldn't see. Ideally, these would come as "whispers" (ala IRC or slack's slash -command api); messages in the channel that only the user can see. But since it seems slack doesn't support this, a dm would be most appropriate. ... I'm still trying to figure out how to send a dm, but the Hubot documentation indicates to me that reply is supposed to.

PR #141, opened 2015-01-11
use JSON.stringify to print Error object

shokai:
## bug
I could not read error's detail, because it is using `toString()`.

```
[Sun Jan 11 2015 16:51:34 GMT+0900 (JST)] ERROR Received error [object Object]
```

## fixed

use JSON.stringify to print error object.
```
[Sun Jan 11 2015 16:59:42 GMT+0900 (JST)] ERROR Received error {"code":-1,"msg":"slow down, too many messages..."}
```

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -40,7 +40,7 @@ class SlackBot extends Adapter
     @client.login()
 
   error: (error) =>
-    @robot.logger.error "Received error #{error.toString()}"
+    @robot.logger.error "Received error #{JSON.stringify error}"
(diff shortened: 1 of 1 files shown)


paulhammond:
Thanks for noticing this and fixing it!

evansolomon:
Just FYI if native `Error` objects are ever passed in here (I don't actually know if they are at the moment) this will probably end up hiding the message.

```js
var e = new Error('whatever')
console.log(e.toString())
// Error: whatever
console.log(JSON.stringify(e))
// {}
```

PR #148, opened 2015-01-31
Add back support for Slack attachments

paulhammond:
This pull request reimplements slack attachment support in Hubot.

It does so using an as-yet undocumented Slack API, where chat.postMessage will now accept an `as_user` boolean argument. Even though this API has been deployed to our production servers, **_it should be considered provisional and the details may change_**. We'll be documenting it on api.slack.com and confirm here when we're confident that the design is right.

I believe this is compatible with both scripts that used hubot-slack v2's attachment support, and scripts that use @inkel's [hubot-slack-attachment](inkel/hubot-slack-attachment) script. But, I don't have many test cases, so I may have missed something.

I'm looking for feedback on whether this approach makes sense, and bug reports where the attachment support doesn't work with existing scripts.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -37,6 +37,9 @@ class SlackBot extends Adapter
     @client.on 'userChange', @.userChange
     @robot.brain.on 'loaded', @.brainLoaded
 
+    @robot.on 'slack-attachment', @.customMessage
+    @robot.on 'slack.attachment', @.customMessage
+
     # Start logging in
     @client.login()
 
@@ -245,4 +248,33 @@ class SlackBot extends Adapter
     channel = @client.getChannelGroupOrDMByName envelope.room
     channel.setTopic strings.join "\n"
 
+  customMessage: (data) =>
+
+    channelName = if data.channel
+      data.channel
+    else if data.message.envelope
+      data.message.envelope.room
+    else data.message.room
+
+    channel = @client.getChannelGroupOrDMByName channelName
+    return unless channel
+
+    msg = {}
+    msg.attachments = data.attachments || data.content
+    msg.attachments = [msg.attachments] unless Array.isArray msg.attachments
+
+    msg.text = data.text
+
+    if data.username && data.username != robot.name
+      msg.as_user = false
+      if data.icon_url?
+        msg.icon_url = data.icon_url
+      else if data.icon_emoji?
+        msg.icon_emoji = data.icon_emoji
+    else
+      msg.as_user = true
+
+    channel.postMessage msg
+
+# Export class for unit tests
 module.exports = SlackBot
--- a/package.json
+++ b/package.json
@@ -67,6 +67,6 @@
     "test": "mocha --compilers coffee:coffee-script/register --reporter spec"
   },
   "dependencies": {
-    "slack-client": "~1.2.0"
+    "slack-client": "~1.4.0"
   }
 }


paulhammond:
I should add, this relies on slackhq/node-slack-client#15 being merged before it will work...

src/slack.coffee
```diff
@@ -242,4 +245,29 @@ class SlackBot extends Adapter
     channel = @client.getChannelGroupOrDMByName envelope.room
     channel.setTopic strings.join "\n"
 
+  customMessage: (data) =>
+    channel = @client.getChannelGroupOrDMByName data.message.envelope.room
+
+    msg = []
```
evansolomon: Should this be an object instead of an array?  Below `msg.attachments` is set, which actually does work as "expected" even though it's an array (because arrays are objects) but it is pretty unusual.

src/slack.coffee
```diff
@@ -242,4 +245,29 @@ class SlackBot extends Adapter
...
+    data.attachments = data.content if data.content and not data.attachments
+    data.attachments = [data.attachments] unless Array.isArray data.attachments
+
+    msg.attachments = []
+    for item in data.attachments
+      msg.attachments.push item
```
evansolomon: I think `msg.attachments = data.attachments` does the same thing as these 3 lines

src/slack.coffee
```diff
@@ -242,4 +245,29 @@ class SlackBot extends Adapter
     channel = @client.getChannelGroupOrDMByName envelope.room
     channel.setTopic strings.join "\n"
 
+  customMessage: (data) =>
+    channel = @client.getChannelGroupOrDMByName data.message.envelope.room
+
+    msg = []
+    data.attachments = data.content if data.content and not data.attachments
+    data.attachments = [data.attachments] unless Array.isArray data.attachments
```
kballard: It might be nicer to not mutate the `data` object. I don't know if clients that post objects to `slack-attachment` or `slack.attachment` ever expect the object to remain un-mutated, but it just seems unnecessary.

kballard:
I'm not entirely clear on what `as_user` is actually doing here.

src/slack.coffee
```diff
@@ -242,4 +245,29 @@ class SlackBot extends Adapter
...
+      msg.attachments.push item
+
+    msg.text = data.text
+
+    if data.username && data.username != robot.name
+      msg.as_user = false
+      if data.icon_url?
+        msg.icon_url = data.icon_url
+      else if data.icon_emoji?
+        msg.icon_emoji = data.icon_emoji
```
kballard: You don't seem to actually be adding `data.username` to the `msg` here. Surely that's necessary or it won't know what name to post as.

inkel: Unless it's getting the user name from the authentication token, which might make sense, but explicitly setting the username makes even more sense, IMO.

src/slack.coffee
inkel: API aren't explicitly clear on the format, but everything suggest that an object is needed, so I think @evansolomon is right.

src/slack.coffee
inkel: Also instead of creating `data.attachments` to later copy it into `msg` in the next three lines, why not directly working on `msg.attachments`.

inkel:
@paulhammond this is really good news, that attachments support is enabled back in the main library. My only concern is relying on an undocumented feature.

podviaznikov:
would love to see it merged into master

omega:
A general feedback, based on how we used slack attachments in hubot in the old version is that we changed the adapter `send` method to treat strings and objects differently. Objects would be sent trough to `customMessage`, and treated as attachments. This allowed us to have our scripts send attachments for slack, and HTML messages for xmpp. It might not be everyones cup of tea, but something like that really made it easier for us to write scripts working with multiple adapters.

(related note is that our internal scripts often reply with an object that will return different representation for different adapters)

src/slack.coffee
ssaavedra: If you want scripts to be compatible with @inkel's hubot-slack-attachment script, you will want to use `data.channel || data.message.room`. [hubot-bitbucket-pr][hbp] for example will not work with your current patch (@paulhammond I have sent you a patch).

  [hbp]: https://github.com/hubot-scripts/hubot-bitbucket-pr

src/slack.coffee
paulhammond: Yep! Thanks for pointing this out, it's fixed in 471e71dea0acdf000f1fb0dfa621156d61148e8e

src/slack.coffee
paulhammond: I've just updated the pull request in 471e71dea0acdf000f1fb0dfa621156d61148e8e to fix all of these issues. Thanks for pointing them out!

src/slack.coffee
paulhammond: `as_user` causes the message to get the name from the authentication token.

This is because behind the scenes `as_user` causes the message to be sent as a [standard message](https://api.slack.com/events/message) instead of a [bot_message](https://api.slack.com/events/message/bot_message), which means the message is correctly attributed to the bot (among many other things).

inkel:
@paulhammond :raised_hands: 

Does this mean that `hubot-slack-attachment` can be deprecated or are is it still compatible?

Either way, this is great news :smiley_cat:

scott2449:
@inkel @paulhammond is there a way to set fallback? When I have notifications set the message is "NO FALLBACK DEFINED" no matter what I add to the payload.

inkel:
@scott2449 could you paste a code snippet of the payload you're using? I could use that to check how am I doing it in [`hubot-slack-attachment`](https://github.com/inkel/hubot-slack-attachment). Thanks!

scott2449:
I tried what you said in the docks and set fallback as such, maybe it's not carried to the PR:

```
robot.emit 'slack-attachment',
              channel: room
              fallback: "*#{issue}* _(#{url})_\nstatus: #{item.fromString} -> #{item.toString}"
              content:
                color: "d96b38",
                fields: [{
                    title: "Status Change",
                    value: "#{item.fromString} -> #{item.toString}"
                }]
```

scott2449:
also tried setting message

scott2449:
NM I had it outside of content

inkel:
:+1:

PR #154, opened 2015-02-11
Skip messages that we've already seen (based on timestamp)

### LINKED ISSUE #153: Hubot repeats himself upon restart
Hi there!  When I ask Hubot a question and he answers it, that's good.  When he restarts 30 minutes later and answers the old question with an old response, that's bad.

Is this a known issue?  I assume Slack is replaying old content to Hubot upon connection and the bot isn't hashing its handled code into the redis-brain so that it can skip replies.  I could look at working up a patch if I get some positive response here.

Thanks :)

evansolomon:
Attempt to fix #153 

I have tested this a little bit but another set of eyes on it would be helpful

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -7,6 +7,7 @@ Util = require 'util'
 
 class SlackBot extends Adapter
   @MAX_MESSAGE_LENGTH: 4000
+  @LAST_MESSAGE_TIMESTAMP_KEY = 'SlackLastMessageTimestamp'
 
   constructor: (robot) ->
     @robot = robot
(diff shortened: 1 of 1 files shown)


paulhammond:
Sorry to take so long to review this!

This area of our API is not well documented, but fortunately I know how it works behind the scenes. The timestamps that are sent are unique and monotonically increasing _per channel_. This means you could get a message in one channel with a TS of `1425060000.000100` then a split second later get a message in another channel with the TS of `1425060000.000001`. So, if we're going to record the last timestamp seen it has to be per-channel :crying_cat_face:

DEGoodmanWilson:
Just to reiterate Paul's point: You can't dedup based on timestamp alone—you have to look at the channel (room) as well.

johnagan:
I don't believe this is still an issue, but if so the PR is outdated. Closing because it's stale. Feel free to reopen with updates.

PR #162, opened 2015-03-07
Fixes to slack.attachment support

spdustin:
When attempting to emit slack.attachment from a router response in our hubot, I saw the dreaded `ReferenceError: robot is not defined` error in the logs. I updated the customMessage function to check the passed username with `@robot.name` rather than `robot.name`.

Then I noticed by passed name wasn't working (was showing up as 'Bot'). Fixed that too. :wink:

Works great now! Thanks for adding support for attachments back into the adapter!

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -265,7 +265,7 @@ class SlackBot extends Adapter
 
     msg.text = data.text
 
-    if data.username && data.username != robot.name
+    if data.username && data.username != @robot.name
       msg.as_user = false
       if data.icon_url?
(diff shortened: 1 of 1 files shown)


bennor:
Any idea when this will find its way into NPM? Would love to be able to use a custom username and emoji when also using attachments.

jeffbyrnes:
:+1: on getting this out the door!

technicalpickles:
cc https://github.com/slackhq/hubot-slack/issues/228 for tracking getting this released

PR #163, opened 2015-03-12
Add section for channel notifications

stonith:
Added a section to clarify how to configure hubot scripts to notify specific channels.


paulhammond:
Thanks for this! This has tripped a few people up, I'm glad we've got it in the docs.

PR #164, opened 2015-03-15
Update README.md

ricardopolo:
Missing attribute. If not specified you get the error `Run this command from an app folder or specify which app to use with --app'


evansolomon:
I think all these commands assume you're in the app directory.  I don't use heroku much, but at the very least the `git push` command assumes you are in the app's directory.  And since it's pushing to a remote created by heroku, I assume `heroku create` is the same.

PR #168, opened 2015-03-18
Allow sending DMs to arbitrary users

jtatum:
This change uses the new callbacks available in node-slack-client. If
robot.send is called like:

robot.send {room: 'someuser'}, 'message'

If someuser is a user and there is no existing DM, it will use openDM to
establish a new private message with the user and send the message in the
callback.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -197,6 +197,12 @@ class SlackBot extends Adapter
   send: (envelope, messages...) ->
     channel = @client.getChannelGroupOrDMByName envelope.room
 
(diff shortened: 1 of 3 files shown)


jtatum:
Note the travis build fails because of dependencies on 0.8, not because of a failing test. I've opened #169 to address this.

paulhammond:
Awesome! Thanks for sharing this (and dealing with our travis needing to be updated)!

technicalpickles:
cc https://github.com/slackhq/hubot-slack/issues/228 for tracking getting this released

PR #182, opened 2015-04-16
SlackClient may not give any arguments if webscoket emit error

iwege:
Relate to slack client https://github.com/slackhq/node-slack-client/blob/master/src/client.coffee#L123-L125

```
 @ws.on 'error', =>
        # TODO: Reconnect?
        @emit 'error'
```

The slack client doesn't pass websocket error in this event.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -44,10 +44,15 @@ class SlackBot extends Adapter
     @client.login()
 
   error: (error) =>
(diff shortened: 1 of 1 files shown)


evansolomon:
Would it make more sense to just pass the websocket error?

iwege:
@evansolomon  yes, I have tried to pass the websocket error to this function in local and it works fine.

PR #190, opened 2015-05-12
Update README.md

davidkassa:
Update to use addons:create instead of :add and switched to rediscloud which offers more free space and is more consistent with Hubot main docs.


mjsz:
Merged the incorrect command in https://github.com/slackhq/hubot-slack/pull/237, and noted that we should update the account as well. Coming soon.

PR #193, opened 2015-05-19
Add support for @mentioning users in hubot message

kdawgwilk:
Feature from #132 

Not sure if this is the right way to go about adding this feature but I did it for our company bot and figured others would like the feature as well.  I implemented in the send method so that you only need to prefix usernames with @ in your scripts and the when it goes through the adapter's send method the `@username` gets replaced with `<@UXXXXX>` to mention a user.  If the the user does not exist then the `@username` gets sent without actually mentioning or sending notification.  

FIXED:
Also I had a weird scoping issue with `@client` (I'm not super familiar with coffeescript) but solved it by making a copy (eg. `client = @client`) if anyone knows a better way to solve that I would be happy to change it

PR #195, opened 2015-05-21
Reply to people with @

### LINKED ISSUE #132: Allow hubot to mention users
Today you can do this in a script to cause a mention:

    msg.send "Hey <@U02871NCZ>"

Where `U02871NCZ` is the internal user id of the slack user you want to mention. But... this isn't exactly intuitive, and is unlikely to be supported by most hubot scripts. We should automatically convert `@paulhammond` to `<@U02871NCZ>` in our send function.

While we're at it, we might want to do the same for `@channel` to `<!channel>` (and group, and everyone).

### LINKED ISSUE #167: Reply should mention user
Hi and thanks for great plugin!

Currently, `msg.reply` will write just `username_to_reply_to:` which is only highlighted in Slack by default. Instead, hubot-slack should use mention name of user.

I've tried just prefixing it with `@`, but it seems like Slack handles mentions differently, so not sure how to implement this.

hkdobrev:
The `@` mention is the canonical in Slack.

Closes #193, fixes #132, fixes #167.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -203,6 +203,14 @@ class SlackBot extends Adapter
     for msg in messages
       continue if msg.length < SlackBot.MIN_MESSAGE_LENGTH
 
+      # Replace @username with <@UXXXXX> for mentioning users
+      msg = msg.replace /(?:^@| @)([A-z]+)/gm, (match, p1) =>
+        try
+          user_id = @client.getUserByName(p1).id
+          match = ' <@' + user_id + '>'
+        catch
+          match = match
+
       @robot.logger.debug "Sending to #{envelope.room}: #{msg}"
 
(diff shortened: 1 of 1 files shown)


hkdobrev:
Hey @evansolomon I've merged my changes with #193 so everything works correctly and it closes a number of issues.

/cc @kdawgwilk

kdawgwilk:
I added support for mentions like `@channel`, `@everyone`, `@group` in #193

tedsteen:
why does this exist when #193 is solving the same thing and is more complete

hkdobrev:
@tedsteen I've built upon #193 and then @kdawgwilk copied everything into his PR and added some more goodies. I'm closing this now. Thank for the reminder.

tedsteen:
sweet :+1: less confusing for new people trying to find their way to the source of the problem/solution

PR #200, opened 2015-06-09
Change "add" to "create" when adding addon to heroku

bognix:
Method add is deprecated in Heroku Toolbelt


tinystride:
+1

```
$ heroku addons:add redistogo:nano
=> WARNING: `heroku addons:add` has been deprecated. Please use `heroku addons:create` instead.
```

mjsz:
Merged (part of) this change in https://github.com/slackhq/hubot-slack/pull/237. Noted that we should update the account as well.

Issue #127, opened 2014-12-17
SlackBot should reconnect instead of terminating the process


kballard:
When the slack connection closes, or when an error occurs (which `SlackBot` currently handles by terminating the process after 1 second), the adapter should reconnect after a short timeout instead of terminating the process.

For precedent, the bundled Campfire adapter reconnects after 5 seconds when its streaming HTTP connection closes.

michael-hopkins:
I believe this is the cause to why my hubot is shutting down on boot. We have a number of scripts all of which have been fully documented. Hubot-slack connects, and then goes to parse all the `hubot help` contents. I believe it takes longer than 10 seconds to go from `bin/hubot - a slack` to idle. I'll get a log tonight and see if that's correct

evansolomon:
@michael-hopkins how many scripts are we talking about?  Taking 10 seconds to parse help content seems *really* surprising, that's many orders of magnitude more time that I would expect.

Also just fyi the command you pasted, `bin/hubot - a slack` has a typo.  There should not be a space between the `-` and the `a`.

michael-hopkins:
Typo edited (mobile). I agree that it seems like a long time. I'll get some logs and see what's up

michael-hopkins:
@evansolomon looked through the logs. On a solid bootup (no errors, and proper), it takes 7 seconds. I happened to still have the logs from when i was having issues. It looks like there was just a slight delay getting the storage from rediscloud. But 3 seconds doesn't leave a whole lot of time for a web service to be a little slow.

ShadowRAM:
I think I am having the same or similar problem.  Hubot runs fine with the irc adapter but it has some lag so I switched to slack and it can't stay up for more than a few hours.  I have two bots and they both wind up doing this after a while:

[Sun Feb 01 2015 23:39:12 GMT-0500 (EST)] ERROR Last pong is too old: 14.997
[Sun Feb 01 2015 23:39:12 GMT-0500 (EST)] INFO Reconnecting in 1000ms
[Sun Feb 01 2015 23:39:13 GMT-0500 (EST)] INFO Attempting reconnect
[Sun Feb 01 2015 23:39:13 GMT-0500 (EST)] INFO Connecting...
[Sun Feb 01 2015 23:39:42 GMT-0500 (EST)] INFO Slack client closed

I saw elsewhere that people had success leaving the heroku keepalive script running (even when hosting elsewhere) and it seemed to make it it stay online for a little longer but still dropped the connection.

[...]

PR #215, opened 2015-07-29
If Slack client closes, exit the hubot process

eriklindebratt:
Since Slack client seems to have troubles reconnecting occasionally (e.g. slackhq/hubot-slack#203), this seems like the only way to solve it right now. I'm running Hubot using supervisor and in the case of Slack client dropping its connection it seems to me that having the Hubot process exit and get restarted from e.g. supervisor is the way to go.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -13,14 +13,14 @@ class SlackBot extends Adapter
     @robot = robot
 
   run: ->
-    exit_process_on_disconnect = !!process.env.HUBOT_SLACK_EXIT_ON_DISCONNECT
+    exitProcessOnDisconnect = !!process.env.HUBOT_SLACK_EXIT_ON_DISCONNECT
 
     # Take our options from the environment, and set otherwise suitable defaults
     options =
       token: process.env.HUBOT_SLACK_TOKEN
-      autoReconnect: !exit_process_on_disconnect
+      autoReconnect: !exitProcessOnDisconnect
       autoMark: true
-      exitOnDisconnect: exit_process_on_disconnect
+      exitOnDisconnect: exitProcessOnDisconnect
 
     return @robot.logger.error "No services token provided to Hubot" unless options.token
     return @robot.logger.error "v2 services token provided, please follow the upgrade instructions" unless (options.token.substring(0, 5) in ['xoxb-', 'xoxp-'])
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -13,11 +13,14 @@ class SlackBot extends Adapter
     @robot = robot
 
   run: ->
+    exit_process_on_disconnect = !!process.env.HUBOT_SLACK_EXIT_ON_DISCONNECT
+
     # Take our options from the environment, and set otherwise suitable defaults
     options =
       token: process.env.HUBOT_SLACK_TOKEN
-      autoReconnect: false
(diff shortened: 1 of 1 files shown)


farski:
Would this be better as an optional behavior change?

KenPowers:
I'd like to see this merged but I agree with @farski that it should be behind an option.

eriklindebratt:
@farski @KenPowers Yep, I agree. Would you prefer something like an env var (e.g. `HUBOT_SLACK_EXIT_ON_DISCONNECT=true bin/hubot --adapter slack`)?

KenPowers:
Either that or if hubot could support [`subarg`](https://github.com/substack/subarg)s then we could have:

```
bin/hubot --adapter slack [ --exit-on-disconnect ]
```

But I think environment variables are more likely.

christopherRiddersater:
Sorry for jumping in like this.
Behind an option would be nice and as @KenPowers points out, a subarg would be nice but either would be a great improvement.

eriklindebratt:
@farski @KenPowers @christopherRiddersater I added a check for an environment variable `HUBOT_SLACK_EXIT_ON_DISCONNECT` and that is used as a flag in `slack.coffee` on whether the Hubot process should exit when the Slack client disconnects. Thoughts?

farski:
@eriklindebratt Looks reasonable to me. Only thing is `exit_process_on_disconnect` should probably be `exitProcessOnDisconnect` by convention.

eriklindebratt:
@farski Ah missed that one, thanks.

christopherRiddersater:
Looks good to me.

spoon99:
+1

bradamson:
+1  This would help my monitoring for sure.  My hubot tends to not connect back, so it sits there not doing anything.

evansolomon:
Is this issue pretty much the opposite of #127?

bradamson:
Seems so, but is configurable on what behavior you want.  If you want it to try reconnecting, leave it as is.  If you want it to exit so supervisor/script/equivalent can detect it died and spin Hubot back up, go for it and turn the option on.

KenPowers:
The behavior I was observing is that my hubot would never reconnect.

bradamson:
Same.  Mine never tries reconnecting (seemingly), so I'd rather let it exit and respin it back up.

ivarrian:
Same as @KenPowers . The process is up but the last log I see is `INFO Slack client closed, waiting for reconnect` , but it doesn't ever reconnect. I have setup foreverjs to restart the process when it dies, but it actually sticks around doing nothing.

KenPowers:
That said, the preferable solution for the long term would be to figure out why hubot isn't reconnecting.

bradamson:
Agreed.

evansolomon:
Seems reasonable to me

christopherRiddersater:
Yes, I agree about the long term solution and I think that the short time solution may prove useful for other scenarios in the future as well.

mjsz:
This looks reasonable as a stopgap, thanks all. I will plan to merge and do a patch release with this included later today.

eriklindebratt:
Sweet @mjsz, thanks!

christopherRiddersater:
@mjsz Great! Many thanks!

MrJoy:
Any notion of when a new release that includes this will be pushed to NPM?

PR #217, opened 2015-08-08
Add edited message handling. Should fix #143.

### LINKED ISSUE #143: Get notified of user editing message
I couldn't find an issue for this and maybe I can do it already but how can the bot know when a user has edited their message. For example someone types a command which has a spelling error in to:

hubot iamge me foobar
::hubot does not respond::
The user then edits their message to say:

hubot image me foobar
::hubot still doesn't respond::

It would be nice if the hubot script could get least get a different message type when the user performs an edit.

pasine:
Edited messages get into the if statement at row 100 in slack.coffee

```
// msg.hidden = true; msg.user = null;

if msg.hidden or (not msg.text and not msg.attachments) 
              or msg.subtype is 'bot_message'
              or not msg.user or not channel
```

Even the content of the message is different, as it has a `message` property with the content of the edited message.

```
  type: 'message',
  message:
   { type: 'message',
     user: 'U0XXXU',
     text: 'Hello World',
     edited: { user: 'U0XXXU', ts: '1438982543.000000' },
     ts: '1438982539.000028' },
  subtype: 'message_changed',
```

I add a subtype statement to handle it as a regular `SlackTextMessage` instead of a `SlackRawMessage`.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -115,6 +115,20 @@ class SlackBot extends Adapter
       if msg.subtype is 'bot_message'
         @robot.logger.debug "Received bot message: '#{text}' in channel: #{channel?.name}, from: #{user?.name}"
         @receive new SlackBotMessage user, text, rawText, msg
+
+      else if msg.subtype is 'message_changed'
+        rawText = msg.getBody()
+
+        # The content of the edited message
+        editedMessage = msg.message
+
+        text = editedMessage.text
+        user = @robot.brain.userForId editedMessage.user
+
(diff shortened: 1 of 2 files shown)


DEGoodmanWilson:
I'm not convinced that edited messages should be replayed to the Hubot scripts like this. That looks like a recipe for trouble. Indeed, most other Slack integrations/apps ignore edited messages, requiring instead explicitly retyping the corrected message. For that reason, I'm tempted to close this PR unmarked. But I'd like to hear your thoughts on the matter before I do.

johnagan:
I'm also 👎 on this. The user has the option to watch this if it's important to them. I'm open to hearing others' views here, but closing until then.

PR #226, opened 2015-09-07
moves hubot out of devDependencies and into dependencies

### LINKED ISSUE #225: Hubt should be a dependency rather than a devDependency
Noticed that `hubot` wasn't installed when only installing production deps:

https://github.com/slackhq/hubot-slack/blob/master/package.json#L60

This should be up in the `dependencies` rather than the `devDependencies`

ajwhite:
This fixes #225.

`hubot` was in the `devDependencies` rather than the `dependencies`. When someone installs only the production dependencies (`npm install --production`), this core dependency is not included.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,8 +56,7 @@
     "should": "~2.0.2",
     "mocha": "~1.13.0",
     "grunt-contrib-watch": "~0.5.3",
-    "grunt-shell": "~0.5.0",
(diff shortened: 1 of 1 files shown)


ajwhite:
Closing as per resolution in https://github.com/slackhq/hubot-slack/issues/225#issuecomment-139017227

PR #230, opened 2015-09-17
robot.emit 'error' for slack client errors

technicalpickles:
I'm debugging an error I received overnight that crashed my hubot instance over at https://github.com/github/hubot-slack/pull/1 . Digging into it a bit more, I found that the adapter's error handler is exiting when the error happens.

This PR updates the adapter to not exit, and to use [hubot's error handling](https://hubot.github.com/docs/scripting/#error-handling) instead of logging it (the error handling logs by default).

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -46,10 +46,7 @@ class SlackBot extends Adapter
   error: (error) =>
     return @robot.logger.warning "Received rate limiting error #{JSON.stringify error}" if error.code == -1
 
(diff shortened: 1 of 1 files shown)


technicalpickles:
Looks like this has my changes from https://github.com/slackhq/hubot-slack/pull/229 in it too. I can rebase this PR with only the emit changes.

technicalpickles:
I've cherry-picked the changes as I mentioned.

PR #233, opened 2015-09-26
Support attachments using adapter.send

technicalpickles:
I'm considering this more a proof of concept, rather than something that should be merged. I was noodling about https://github.com/github/hubot/issues/1036, and I got to thinking, maybe `res.send` should support being able to send attachments, rather than adding more functions to the adapter/response API?

Example script:

```
  robot.respond /PING$/i, (msg) ->
    msg.send {attachment: {color: 'good', text: "PONG"}}
```

There is one pretty glaring problem though: it breaks the adapter interface.https://github.com/github/hubot/blob/master/src/adapter.coffee#L9-L15 says:

``` coffeescript
  # Public: Raw method for sending data back to the chat source. Extend this.
  #
  # envelope - A Object with message, room and user details.
  # strings  - One or more Strings for each message to send.
  #
  # Returns nothing.
  send: (envelope, strings...) ->
```

More importantly than some theoretical API breakage, this could be super confusing for script authors and users. You can just `msg.send`, and it's not clear what would happen. If you wanted to use `msg.send` with an attachment, it'd probably look super weird on any other adapter.

Like I said, more of an idea exploration, and probably a good case for extending adapters/responses as described in https://github.com/github/hubot/issues/1036

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -201,44 +201,49 @@ class SlackBot extends Adapter
       return
 
     for msg in messages
-      continue if msg.length < SlackBot.MIN_MESSAGE_LENGTH
+      if typeof(msg) is 'string'
+        continue if msg.length < SlackBot.MIN_MESSAGE_LENGTH
 
-      @robot.logger.debug "Sending to #{envelope.room}: #{msg}"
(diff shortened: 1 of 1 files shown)


thetrav:
uploading would be very useful, specifically creating snippets is something I want to do frequently

PR #235, opened 2015-09-27
Send long messages as attachments with text

technicalpickles:
This prevents very long messages from flooding channels. It's triggered
by 700 characters or 5 lines of text. This was chosen because of the
attachment documentation at https://api.slack.com/docs/attachments saying:

> The content will automatically collapse if it contains 700+ characters or 5+ linebreaks, and will display a "Show more..." link to expand the content.

This would fix https://github.com/slackhq/hubot-slack/issues/17

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -206,7 +206,10 @@ class SlackBot extends Adapter
       @robot.logger.debug "Sending to #{envelope.room}: #{msg}"
 
       if msg.length <= SlackBot.MAX_MESSAGE_LENGTH
-        channel.send msg
+        if msg.length >= 700 || msg.split(/\n/).length > 5
+          @customMessage channel: envelope.room, content: {text: msg}
+        else
+          channel.send msg
 
       # If message is greater than MAX_MESSAGE_LENGTH, split it into multiple messages
       else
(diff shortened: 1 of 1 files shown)


technicalpickles:
Sending multiple submessages actually ends up looking a little weird, since there's multiple "show more":



I wonder if it'd be better to just truncate, and log a warning.

I've also asked Slack support about confirming the maximum message length.

willdurand:
Hi! Any news in this?

technicalpickles:
I haven't received any feedback on this yet. There are failing tests, and they are complex enough that I didn't want to dig into them unless this is in the right direction.

sheluchin:
>I wonder if it'd be better to just truncate, and log a warning.
I don't think it would be better. That warning would say nothing about the contents of the truncated message and finding the message would be difficult after seeing a warning. The warning is basically not actionable. Attaching long messages as files seems like the most logical approach. At least that way we could see the contents of the message and act on it, if necessary. I would like to see this effort move forward.

sheluchin:
@technicalpickles: do you know if the problem changes at all with SDK v3 introduced in #309?

DEGoodmanWilson:
We're looking into it @sheluchin :D

aoberoi:
i don't quite understand what flooding means in this context. Slack will fold up long messages so that a single long message cannot take over the entire display. From my point of view, that is preventing flooding. It seems like this PR is meant to circumvent that folding of long messages, so it would only increase flooding.

Closing for now. If you can help me understand the problem this PR aims to solve, i'll reopen it.

PR #236, opened 2015-09-29
Always use chat.postMessage with attachments for sending

technicalpickles:
This accomplishes a few things:

* allow scripts to use Slack formatting
* long messages are automatically collapsed with 'show more'
* avoid hitting API rate limiting when sending long messages that get
  split up

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -203,14 +203,19 @@ class SlackBot extends Adapter
     for msg in messages
       continue if msg.length < SlackBot.MIN_MESSAGE_LENGTH
 
-      @robot.logger.debug "Sending to #{envelope.room}: #{msg}"
-
       if msg.length > SlackBot.MAX_MESSAGE_LENGTH
         @robot.logger.warning "Tried to send #{msg.length} character message, which is bigger than Slack's #{SlackBot.MAX_MESSAGE_LENGTH} maximum: truncating it"
 
         msg = msg.substring(0, SlackBot.MAX_MESSAGE_LENGTH)
 
-      @customMessage channel: envelope.room, attachments: {text: msg, mrkdwn_in: ['text']}
+      attachment = if msg.match(/^https?:\/\/\S+(?:jpg|png|gif)$/)
+        @robot.logger.debug "Sending to #{envelope.room} as image: #{msg}"
+        {image_url: msg, fallback: msg}
+      else
+        @robot.logger.debug "Sending to #{envelope.room}: #{msg}"
+        {text: msg, mrkdwn_in: ['text']}
+
+      @customMessage channel: envelope.room, attachments: attachment
 
   reply: (envelope, messages...) ->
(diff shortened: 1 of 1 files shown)


technicalpickles:
One downside of splitting up large messages: If you use triple-backticks to do preformatted text, splitting up the message means that it'd start with ```, but the ending ``` would be in a following message, so they wouldn't be formatted.

Could check explicitly for messages starting like that, and make sure to wrap the split up messages?

technicalpickles:
> Could check explicitly for messages starting like that, and make sure to wrap the split up messages?

I thought about this a bit, and now I'm not so sure. The logic to split up is a bit complicated, so trying to parse out triple back ticks would make it even more so. I took a pass at doing that, but wasn't having much luck.

I think it's a lot simpler to just truncate and log that it happened. The formatting of a large message will be messed up still, but I think it's a bit more predictable. I've updated to do this, which logs a message like:

[Tue Sep 29 2015 17:07:30 GMT-0400 (EDT)] WARNING Tried to send 8662 character message, which is bigger than Slack's 8000 maximum: truncating it

I'd be good to log where this happened, but that might be a bit involved.

technicalpickles:
I've been using this in production for a few weeks, and ran into something pretty problematic: If you send an image URL, it doesn't get unfurled:



One work around would be to try to detect an image URL, and set `image_url` instead.

technicalpickles:
> If you send an image URL, it doesn't get unfurled:

I think this is a case for having something on hubot's `Response` like `sendImage` that was proposed in https://github.com/github/hubot/pull/996 . Parsing the text to figure out what attachment to send doesn't feel great, when it's pretty clear the intent is to send an image.

technicalpickles:
Looking at https://api.slack.com/docs/unfurling and https://api.slack.com/methods/chat.postMessage , it looks like there's a `unfurl_media` option which defaults to `false`. I'm not sure what that does on the backend, and if it's a safe thing to default to :sweat: 

It's about here I think yet again about `res.sendImage` that takes URLs to send.

technicalpickles:
> Looking at https://api.slack.com/docs/unfurling and https://api.slack.com/methods/chat.postMessage , it looks like there's a unfurl_media option which defaults to false

I misread that, it's enabled by default. I got confused by the example they gave, which was `false` which was to disable it, as in, it's enabled by default.

technicalpickles:
I am looking for feedback if this is a good approach or not, or if the downsides are outweighed by the benefits.

The tests are failing, but the code works (we've been using it in production since I made this PR). I hadn't been able to work out the failures yet, but I was also hoping for some comments one way or the other.

technicalpickles:
https://github.com/slackhq/hubot-slack/pull/314 sends attachments now, but two problems with it compared to this branch:

* the message comes as a bot user
* text isn't formatted in the response

Consider sending a normal message, vs attchment for the same content:





cc @johnagan

PR #242, opened 2015-10-11
Correctly @-mention users when using msg.reply

### LINKED ISSUE #167: Reply should mention user

bojanrajkovic:
When using `msg.reply`, users would have to have their Slack nickname as a highlight word to receive notifications, because Hubot wasn't sending the special formatting to generate an @-reply. This fixes it so that Hubot generates correct @-replies.

Fixes #167.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -246,7 +246,7 @@ class SlackBot extends Adapter
 
     for msg in messages
       # TODO: Don't prefix username if replying in DM
-      @send envelope, "#{envelope.user.name}: #{msg}"
+      @send envelope, "<@#{envelope.user.id}>: #{msg}"
 
   topic: (envelope, strings...) ->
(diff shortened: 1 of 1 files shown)


bojanrajkovic:
This looks like it should fix #195

Floby:
I would like this :3

bojanrajkovic:
It's been annoying me for the longest time. Seems so basic. >_>

Floby:
I see this has been merged, any chance of it getting published soon? =) pretty please

PR #246, opened 2015-10-21
impr: add support for mentioning users in hubot messages

### LINKED ISSUE #132: Allow hubot to mention users

QuentinFchx:
Original work from @kdawgwilk in #193

Closes #132

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -8,6 +8,7 @@ Util = require 'util'
 class SlackBot extends Adapter
   @MAX_MESSAGE_LENGTH: 4000
   @MIN_MESSAGE_LENGTH: 1
+  @RESERVED_KEYWORDS: ['channel','group','everyone','here']
 
   constructor: (robot) ->
     @robot = robot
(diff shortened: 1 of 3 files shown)


QuentinFchx:
Who is modifying class variables during tests without resetting them ? :O :O :O

QuentinFchx:
@mjsz this is an updated work of what has been done in #193 around mentioning users.

technicalpickles:
Does this work as expected? Without https://github.com/slackhq/hubot-slack/pull/236 it'd be posting back using the rtm API, which I've found doesn't support slack formatted links.

PR #248, opened 2015-11-05
Change back quotes as Slack tends to change them on user input.

szilardhuber:
Some hubot plugins (like [hubot-trello](https://github.com/hubot-scripts/hubot-trello) match on quotes and they fail to work together with Slack as when a user enters ' Slack changes it to ‘ and the same happens to double quotes. 

I think it is best to replace those characters in the removeFormatting method but this is just a naive first try from me so please point out if I made any mistakes or the whole assumption is wrong.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -19,6 +19,7 @@ class SlackBot extends Adapter
       token: process.env.HUBOT_SLACK_TOKEN
       autoReconnect: true
       autoMark: true
+      replaceQuotes: process.env.HUBOT_SLACK_QUOTES or false
 
     return @robot.logger.error "No services token provided to Hubot" unless options.token
     return @robot.logger.error "v2 services token provided, please follow the upgrade instructions" unless (options.token.substring(0, 5) in ['xoxb-', 'xoxp-'])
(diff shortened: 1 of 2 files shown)


evansolomon:
This seems slightly risky but generally reasonable to me. A more conservative approach would be to offer some kind of option/environment variable to disable the behavior. An even more conservative approach would be to use the option/variable to **enable** the behavior (i.e. turn it off by default since, strictly speaking, it's backward incompatible).

szilardhuber:
I added the option but feel uneasy because of the replacePattern right now. I did not want to change the literal approach as I read that this is more efficient than using strings as the pattern but this way I saw no other option than copypasting the literal.

aoberoi:
sorry that this action is **extremely** long overdue.

the maintainers feel that changing the curly quotes to straight quotes  isn't something the adapter needs to solve. this could be a simple one-liner within a script:

```
var text = message.text.replace(\[‘’]\g, '\'').replace(\[“”]\, '"');
```

for anything more advanced, you could use a package like [retext-quotes] (https://www.npmjs.com/package/retext-quotes)

PR #250, opened 2015-11-10
Match slack usernames which have non alpha-numeric characters in them

optical:
The change to allow the adapter to convert @mentions into proper `<@U123>` is excellent - however the matching isn't quite correct.

Currently it'll only match if the username is entirely alphanumeric, however slack usernames can contain any characters excluding whitespace. Within our slack team we have our usernames set as firstname.lastname - which doesn't work with the current implementation. I suspect that many professional slack teams use a similar username system.

Changing the implementation to match all non-whitespace characters doesn't work quite as expected either however, as a common use-case will be to mention usernames and suffix their usename with a colon, eg `@john.smith:`.

This change makes it match the entire word, eg `john.smith:` and look up the user. If it is a match, then we substitute. If no match is made and the trailing character is a non-word, we prune it and try again. We do this until we match a username, exhaust all characters or encounter a word character. This allows scripts to refer to any valid slack username and also append any trailing characters onto them such as colons, eg `@john.smith` or `@john.smith:`

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -214,12 +214,18 @@ class SlackBot extends Adapter
       continue if msg.length < SlackBot.MIN_MESSAGE_LENGTH
 
       # Replace @username with <@UXXXXX> for mentioning users and channels
-      msg = msg.replace /(?:^| )@([\w]+)/gm, (match, p1) =>
-        user = @client.getUserByName p1
+      msg = msg.replace /(?:^| )(@([\S]+))/gm, (match, p1, p2) =>
+        # Greedily match the username, pruning off trailing non-word characters until we get a full match
+        trimmedArgument = p2
+        user = @client.getUserByName trimmedArgument
+        while not user and trimmedArgument.length > 1 and trimmedArgument[trimmedArgument.length - 1].match /\W/
+          trimmedArgument = trimmedArgument.slice 0, trimmedArgument.length - 1
+          user = @client.getUserByName trimmedArgument
+
         if user
-          match = match.replace /@[\w]+/, "<@#{user.id}>"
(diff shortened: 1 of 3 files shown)


DEGoodmanWilson:
Closing this as a duplicate of #250

iancward:
@DEGoodmanWilson I just went looking for a way to get slack IDs from user name mentions, and it looks like you closed this PR as a duplicate of itself.

DEGoodmanWilson:
Hahahah, so I did. Thanks for pointing it out. The issue is, nonetheless fixed ;) I'll update the comment above, it is a duplicate of _something_

DEGoodmanWilson:
Should be fixed since 22 Jul: #310

iancward:
Thanks! So this is just outbound from Hubot to slack? Is there any way to get the slack user ID from an @mention that was sent to the bot?

optical:
@iancward if the mention is a proper slack `@mention`, the actual message will be a string like `Please do something for <@U024BE7LH|ian.ward>`. I don't remember the exact APIs, but from that it'll be possible to lookup the user in hubot by id.

iancward:
@optical when I write the message to the log using `robot.logger.debug`, I just see `hubot: please do something for @ian.ward`, which makes me think it's not actually a link.

DEGoodmanWilson:
That might not be a valid username in the message then. That's curious. Can you open a new issue with concrete steps to reproduce, and maybe a failing test to demonstrate?

iancward:
It looks like the inbound message formatter strips the user id from the message, leaving only the `@` mention in the message: https://github.com/slackapi/hubot-slack/blob/89625873b21dc70337737494cb7b66f6ea5c2056/test/formatter.coffee#L23-L25

Issue #256, opened 2015-11-21
Sending customMessage() with attachment payload via DM when no previous conversation


ehammersley:
Very similar to #204, which was resolved via #168 in cases where adapter.send is in use.  Based on what I can understand it checks for an existing IM.id with the user, and if one doesn't exist it opens a new IM.

However, I had a need for a formatted attachment to be sent via DM.  I ran across and followed the customMessage() method outlined in the comments of #170, which basically involves creating the msgData necessary for an attachment and passing it as data to .customMessage.  However, this only works if a previous conversation existed between the bot and user.  If no previous IM id exists it silently fails.

Is there a way to take the changes added to adapter.send in the #168 merge and apply that to the customMessage() method in the adapter?  Is this even the best way to do it?  Certainly open to suggestions on my approach as well.

Here's an idea of how I'm building the data and sending it.  It works great, so long as it's an existing private DM, otherwise nothing.

```
msgData = {
          channel: msg.envelope.user.name
          attachments: [
            {
              fallback: "Fallback text here.",
              color: "danger",
              title: "Title text here"
              text: "Main body text"
              mrkdwn_in: ["text"]
            }
          ]
        }
        robot.adapter.customMessage msgData
```

evansolomon:
This sounds like it makes sense. Seems like the right approach would be to create a DM instead of returning [here](https://github.com/jtatum/hubot-slack/blob/6a3ceefff159115c61819851b383c8e2406405b6/src/slack.coffee#L259-L260). Is that right?

ehammersley:
That's what I'm thinking, yes.  Something is done very similar to this in master that resolves the issue for adapter.send.  However, the structure of customMessage would make the approach a little different I think.

Here's how it's handled in master for .send.
```
channel = @client.getChannelGroupOrDMByName envelope.room
channel = @client.getChannelGroupOrDMByID(envelope.room) unless channel

if not channel and @client.getUserByName(envelope.room)
  user_id = @client.getUserByName(envelope.room).id
  @client.openDM user_id, =>
    this.send envelope, messages...
  return
```
In customMessage the eval is done on channel, with the return, prior to the message being formatted.  I wonder if one could use the above fix for .send, remove the return after the channel eval, and add a conditional send to the end, checking for the DM and creating if necessary, or posting the message if a regular channel?

I'm happy to test a proposed solution.  I'm not sure I'm good enough of a coder to contribute, but I can give it a whirl if needed.

[...]

PR #261, opened 2015-12-03
Adding new DM creation to customMessage when none previously existed.

ehammersley:
This attempts to extend #168 to customMessages, and was generated in response to and fixes issue #256.

Previously checks for existing DM.id were added to .send in #168, however .customMessage still suffered from the issue and would only send a DM if one previously existed.  This change allows messages to be sent to a user via direct message if no previous direct messages existed when using .customMessage and now aligns behavior with .send.

Note: If there's a better, cleaner, way to accomplish this please let me know.  It's my first crack at contributing to a project.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -39,6 +39,7 @@ class SlackBot extends Adapter
     @client.on 'close', @.clientClose
     @client.on 'message', @.message
     @client.on 'userChange', @.userChange
+    @client.on 'raw_message', @.rawMessage
     @robot.brain.on 'loaded', @.brainLoaded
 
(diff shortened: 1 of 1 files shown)


aoberoi:
sorry for the **extremely late** feedback.

this PR is a bit out of date. we no longer have `SlackRawMessage` in this adapter, instead there is a `rawText` property of `SlackMessage`, which only conditionally exists (see `noRawText` option for `SlackClient`). there are probably bigger differences, but i'm not super familiar with the state of the codebase from this timeframe.

ehammersley:
Close since the PR is way out of date now, and the underlying striation of the original ask is no longer present in the codebase.

PR #262, opened 2015-12-04
Updated README.md initialization steps

etothemanders:
Removed the `npm install hubot-slack step`.  If you change the adapter from the default configuration ('campfire') during `yo hubot` interactive setup (or with the --adapter=slack flag), the hubot-slack module will be installed for you as the final dependency.  Using npm to install it again is redundant.


johnagan:
@DEGoodmanWilson LGTM
@episod may have some feedback

PR #264, opened 2015-12-15
Mark messages not from public channels as 'private'

technicalpickles:
This lets scripts and middleware know if they were said in private channels
 or direct messages, so they can choose to not log or otherwise output
information that was said in the channel.

Here's an example:

``` coffeescript
module.exports = (robot) ->
  robot.respond /PUBLICPING$/i, onlyPublic: true, (msg) ->
    msg.send "PUBLIC PONG"

  robot.listenerMiddleware (context, next, done) ->
    return next() unless context.listener.options?.onlyPublic

    if context.response.message.private
      done()
    else
      next()
```

For prior art, hubot's campfire adapter uses this for [locked messages](https://github.com/github/hubot/blob/master/src/adapters/campfire.coffee#L73), which are messages that aren't logged. It's not documented at the moment though.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -148,15 +148,15 @@ class SlackBot extends Adapter
     # Test for enter/leave messages
     if msg.subtype is 'channel_join' or msg.subtype is 'group_join'
       @robot.logger.debug "#{user.name} has joined #{channel.name}"
-      @receive new EnterMessage user
+      message = new EnterMessage user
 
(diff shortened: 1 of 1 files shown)


johnagan:
This an is interesting idea - unfortunately this PR is stale now since the SDK and Hutbot adapter have evolved. I'm closing this for now, but reopen with updated changes if you still feel strongly about it. Thanks!

PR #279, opened 2016-02-20
hubot-slack-pull

alok-garg:
hubot-slack-pull


l12s:
This seems like a broken PR, so closing it

PR #281, opened 2016-03-03
Hubot Is Typing

RIAlizer:
Added Hubot "Is Typing" feedback
https://github.com/slackhq/hubot-slack/issues/259

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -118,6 +118,9 @@ class SlackBot extends Adapter
 
     channel = @client.getChannelGroupOrDMByID msg.channel if msg.channel
 
+    # send typing feedback
+    @client._send({ type: "typing", channel: msg.channel });
+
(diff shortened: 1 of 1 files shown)


dasilvacontin:
@RIAlizer why did you close it?

dasilvacontin:
Ok, I see.

RIAlizer:
@dasilvacontin The command `_send({ type: "typing")` actually works, but I need to fix tests and add more logic.

dasilvacontin:
But it will be some kind of method you have to manually invoke, I hope? Most common use case is to fake typing when processing a command is gonna take a while.

tlytg456:
@aoberoi Our bot may do some complete work which spends a long time, so we want to give users some feedback while we calculate the results, besides the "xxx is typing" or some other message. We think the "typing" feedback is suitable. But why the project remove this feature?

tlytg456:
@aoberoi Thanks for your patience, I have known that and find the best ways to solve them

PR #291, opened 2016-03-28
Allow commands to be run with alias/name prefix in direct messages

anupdhml:
From direct messages to hubot, commands like `help` work fine when they are not prefixed by anything else. However, people are used to prepending the hubot name, or the alias to run the commands in channels, and if they do the same in DM's, the commands don't work currently. 

This pull request makes command invocation consistent in both channels and DM's (i.e. this enables `hubot help` or `:help` (assuming `HUBOT_ALIAS` is set to `:`), in addition to just `help`)

For reference, other adapters implement similar [functionality](https://github.com/markstory/hubot-xmpp/issues/85#issuecomment-142391307).

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -168,8 +168,8 @@ class SlackBot extends Adapter
       # If this is a DM, pretend it was addressed to us (prepend the robot name).
       # But do it only if the text does not already start with the name, or the alias.
       if msg.getChannelType() == 'DM'
-        nameRegex = "@?#{robot.name}"
-        prefixRegex = if robot.alias then "(#{nameRegex}|#{robot.alias})" else "#{nameRegex}"
+        nameRegex = "@?#{@robot.name}"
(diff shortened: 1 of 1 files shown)


ESPNMichaelKidd:
:+1:

aoberoi:
@anupdhml im sorry for this **extremely** late feedback. i think this is a great idea!

the repo has changed quite a bit, but if you don't mind introducing your changes on top of the latest master, i'd be happy to accept it. essentially i think your code would land here: https://github.com/slackapi/hubot-slack/blob/4229577c1c0723becfe4abe13bb7e0340afc84b2/src/bot.coffee#L178. could you also implement a test?

aoberoi:
still open to this if anyone wants to resolve merge issues

patcon:
I think this can be closed now. Thanks a bunch for the initial effort @anupdhml , and the PR that inspired mine :tada: 
https://github.com/slackapi/hubot-slack/pull/422

PR #293, opened 2016-05-03
Allow processing messages from certain bots

ernesto-jimenez:
Use the env variable `HUBOT_SLACK_BOTS_TO_LISTEN_TO` to set a comma
separated list of bot IDs whose messages should be processed as if they
were from users.

### DIFF
--- a/src/slack.coffee
+++ b/src/slack.coffee
@@ -19,6 +19,7 @@ class SlackBot extends Adapter
     # Take our options from the environment, and set otherwise suitable defaults
     options =
       token: process.env.HUBOT_SLACK_TOKEN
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@DEGoodmanWilson here's what we chatted about yesterday.

How do you run a hubot with a local development version of this package?

I tried with `npm link hubot-slack` pointing to my local fork, but hubot would ignore the registered messages and `@client.bots` was empty.

DEGoodmanWilson:
@ernesto-jimenez we are simply going to allow all bot messages in with #309, mooting this work a bit. You'll still be able to filter on a whitelist, of course, but this change moves the needed functionality out of the SDK and into your implementation.

PR #307, opened 2016-06-02
Add support for emoting

ilikebits:
This supports emoting by replicating the official Slack client's `/me` command.


aoberoi:
@ilikebits: we're not interested in merging this addition. there is already an official API for the functionality you want ([docs](https://api.slack.com/methods/chat.meMessage)). you can invoke this from the web client or rtm client in you're interested.

Issue #62, opened 2014-07-08
upload file


andrewfsmith:
Is there a way to upload a file to slack?
I see how to send messages and links, but I want to upload a local image as an attachment.
Is this possible?

ChrisLundquist:
After taking a look at https://api.slack.com/methods/files.upload
It looks like there is an error "User is bot", which might be why this never made it into the adapter.
I would guess that using a real user's API token, might be a work around.
<img width="665" alt="screen shot 2015-07-09 at 12 31 51 pm" src="https://cloud.githubusercontent.com/assets/181256/8604976/c92e4584-2636-11e5-9592-52097d21c848.png">

victorcoder:
I asked slack support about this and the recommended way to do it is using a non bot user.

technicalpickles:
cc https://github.com/github/hubot/pull/996 where there's some discussion around adding a method to hubot adapters for uploading files

aaronthebaron:
It looks like they've recently changed this policy and now bot users are allowed to upload files: 
![screen shot 2016-01-05 at 2 24 00 pm](https://cloud.githubusercontent.com/assets/3069310/12129375/0dd932fc-b3b8-11e5-9c19-4f32509fca87.png)

santiycr:
+1

jiang-wei:
+1

halkeye:
I feel like https://github.com/slackhq/node-slack-client/pull/56 should be a priority to get merged, the code is written up it just looks like the PR has fallen silent.

halkeye:
@technicalpickles do we have any sort of standard for uploadFile? or do you want me to make up one like I did for emote?

I have a PR in for https://github.com/crossorigin/node-slack-upload/pull/2 which mostly works anyways, I was going to write a stub/fill for response.uploadFile using that for now.

technicalpickles:
@halkeye there isn't. There is discussion in https://github.com/github/hubot/issues/1036 about more 'types'. I think uploadFile makes sense though.

sheluchin:
+1

DEGoodmanWilson:
This will be fixed in https://github.com/slackhq/hubot-slack/pull/309

PR #309, opened 2016-06-16
Updated Logic and Tests for Slack SDK v3 support

johnagan:
This is a first pass at adding support for v3 of the Slack SDK :boom:

The core logic has been updated to use `dataStore` and the test stubs have been modified to support the relocation of data calls.


src/slack.coffee
```diff
@@ -17,33 +15,31 @@ class SlackBot extends Adapter
...
-    options =
+    @options =
+      autoMark: true
+      logLevel: 'error'
+      dataStore: new MemoryDataStore()
       token: process.env.HUBOT_SLACK_TOKEN
       autoReconnect: !exitProcessOnDisconnect
-      autoMark: true
-      exitOnDisconnect: exitProcessOnDisconnect
-      proxyUrl: process.env.https_proxy
```
DEGoodmanWilson: I think the change in the options means this qualifies as a breaking change, and hence will require us to bump the major version number.

DEGoodmanWilson: I also want to verify that the reconnection logic is working.

src/slack.coffee
```diff
@@ -214,21 +205,12 @@ class SlackBot extends Adapter
...
   send: (envelope, messages...) ->
-    channel = @client.getChannelGroupOrDMByName envelope.room
-    channel = @client.getChannelGroupOrDMByID(envelope.room) unless channel
-
-    if not channel and @client.getUserByName(envelope.room)
-      user_id = @client.getUserByName(envelope.room).id
-      @client.openDM user_id, =>
-        this.send envelope, messages...
-      return
-
```
DEGoodmanWilson: Why did this logic get cut? Did I do that?

johnagan: Nope. I cut it because we're send from the client now, rather than [then channel object](https://github.com/slackhq/hubot-slack/pull/309/files/cd61c608b2084759c51c95ed0777208a1e91c9e1#diff-44aca805fd3e634b2ed4f6bc2ccb0749L242). I didn't see where else we were using channel there and figured it could go. 

Very possible I missed something though.

DEGoodmanWilson: It looks like if it doesn't find a valid channel in the envelope, it opens a DM with the sending user. The reason and purpose of this logic isn't totally clear to me, but it is a a bit of functionality we probably shouldn't elide.

src/slack.coffee
```diff
@@ -103,23 +99,28 @@ class SlackBot extends Adapter
...
     # Ignore our own messages
-    return if msg.user == @self.id
+    return if msg.user and msg.user == @id
+
+    channel = @client.dataStore.getChannelGroupOrDMById msg.channel if msg.channel
+    user = @client.dataStore.getUserById msg.user if msg.user
+    user.room ?= msg.channel if channel and user
 
-    channel = @client.getChannelGroupOrDMByID msg.channel if msg.channel
+    rawText = msg.getBody()
```
DEGoodmanWilson: Getting `ERROR TypeError: msg.getBody is not a function` here.

DEGoodmanWilson:
There's another big issue: If you send a message to the bot in a public channel, it responds in a DM—always. The original responded in the same channel.

johnagan:
@DEGoodmanWilson I had to roll back your [package.json](https://github.com/slackhq/hubot-slack/commit/7441fad54c9a7ff290c9c79469277b5e61c594cb#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) commit. Looks like you overwrote it with the version from your `node_modules` folder 😎

src/formatter.coffee
```diff
@@ -0,0 +1,109 @@
...
+      text.push(attachment.fallback)    
+
+    # flatten array
+    text.join('\n')
+
+
+  ### 
+  Recursively replace @username with <@UXXXXX> for mentioning users and channels
+  ###
+  mentions: (text) ->
```
johnagan: @DEGoodmanWilson I just ran across the [`link_names`](https://api.slack.com/methods/chat.postMessage) property on the message object that seems to do this for us. I think we can ditch this whole function.

src/formatter.coffee
```diff
@@ -0,0 +1,109 @@
...
+# https://api.slack.com/docs/formatting
+class SlackFormatter
+
+  constructor: (@dataStore) ->
+
+
+  ###
+  Formats links and ids
+  ###
+  links: (text) ->
```
johnagan: @DEGoodmanWilson I think we can replace this function with [`{parse: true}`](https://api.slack.com/methods/chat.postMessage#formatting)

dnrce:
🎉 👍

CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+  * Better (and automatically enabled) reconnect logic. As in, it actually reconnects automatically at all.
+  * Now you can upload files!
+  * Significantly improved handling of messages with attachments, which is to say, we can deliver them.
+  * Message formatting of links, usernames and channel names is now working far better than it ever did, which is damning with faint praise, but hey.
+  * Long messages are now left for Slack to handle, bless their hearts.
+  * Slack usernames with `.` and `-` are now treated with the respect and dignity due to all usernames.
+  * Messages from bots are no longer filtered out, which is both cool and potentially terrifying, but we should never have silenced the robots in the first place.
+  * Remember how if you tried to hack on this adapter and used `npm link` to plug that into a live bot? And how that didn't work? Yeah? Well now it does. Stupid `instanceof`.
+  * Total refactoring of the functionality, exposing a slightly different interface. So watch out for that.
+  * You can now access the underlying Slack client directly, for when you really need low-level functionality therein.
```
peritus: Is there documentation available on how to do just that ?

johnagan: [`.rtm` or `.web`](https://github.com/slackhq/hubot-slack/blob/09ba54d7ba3ff3993759bdfe7203ea11774c0d60/src/client.coffee#L15) will return the clients.

[`.send`](https://github.com/slackhq/hubot-slack/blob/09ba54d7ba3ff3993759bdfe7203ea11774c0d60/src/client.coffee#L73) is just a shortcut to select the best method.

peritus: Thanks!

ifreecarve:
If there is documentation on how to upload files or send attachments in hubot slack messages, I'm having a lot of time locating it. Here's where I looked:

* https://github.com/slackapi/node-slack-sdk/pull/56
* https://github.com/slackapi/hubot-slack/issues/62
* https://hubot.github.com/docs/scripting/
* https://github.com/slackapi/hubot-slack/pull/309/files (this PR, which mentions that you can "upload" "files" but doesn't include those words anywhere in the diff)

Can someone point me to a brief example of hubot uploading a file to a slack channel?

Treyone:
@ifreecarve did you find out how to upload files ? Same situation here. Thanks !

ifreecarve:
Elsewhere in this thread, it was noted that 
> `.web` or `.rtm` will return the clients.

So I assume that, worst case, you get the native client and access that method directly.  Haven't tried it yet.

Treyone:
Thanks, I'll give it a try !

ifreecarve:
I'm using `robot.adapter.client.web.files.upload` to access the function described here: https://api.slack.com/methods/files.upload

Treyone:
Thanks ! the upload works. The binary files are however displayed as text and unreadable, but this seems to be related to another issue.

ifreecarve:
in `contentOpts`, use `file: fs.createReadStream(filePath)`.  

`content: fs.readFileSync(filePath)` doesn't work with binary

PR #315, opened 2016-07-05
V3 sdk updates writing tests

DEGoodmanWilson:
@johnagan for your perusal.

- Broke existing tests apart for SlackBot and Formatter
- Removed message splitting tests, as I understand we are removing that functionality?
- Reworked stubs to actually work
- Added some logic to sending messages so that empty messages aren't sent (per existing tests read as a spec for functionality)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "url": "https://github.com/slackhq/hubot-slack/issues"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


DEGoodmanWilson:
`Error: Cannot find module '../src/Formatter'`
wat

johnagan:
LGTM 👍

PR #318, opened 2016-07-13
Add proxy support based on new Slack SDK v3

rizkiarm:
This will add a proxy support using wsTransport for ws/wss, proxiedRequestTransport for http/https. 
Proxy url can be configured by setting the environment variable `HUBOT_SLACK_PROXY`.

### DIFF
--- a/index.coffee
+++ b/index.coffee
@@ -1,4 +1,4 @@
 SlackBot = require './src/bot'
 
 exports.use = (robot) ->
-  new SlackBot robot, token: process.env.HUBOT_SLACK_TOKEN
+  new SlackBot robot, token: process.env.HUBOT_SLACK_TOKEN, proxyUrl: process.env.HUBOT_SLACK_PROXY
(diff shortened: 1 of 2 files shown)


aoberoi:
@rizkiarm: thanks for the contribution. i can see how this feature would be useful. would you please add some tests for the new functionality? if you need help with this, let us know and we can take a closer look.

aoberoi:
@rizkiarm ping! i'd still be happy to merge this change, but i need some tests. let me know if you want help with that and if you're still interested in contributing this change. if so, i'll also nee you to sign our CLA.

you might have noticed that the initialization code for `SlackBot` has changed. you should be able to remove your changes in `index.coffee` and [use environment variables to define the options](https://slackapi.github.io/hubot-slack/advanced_usage#customizing-the-rtm-client).

aoberoi:
before landing this, we should update the documentation to show how to use this feature.

aoberoi:
i think this is duplicate of #372, since essentially we're talking about one environment variable either way, and documenting it (and fixing that issue) would solve this one too.

PR #319, opened 2016-07-13
Listen to user_change & brain loaded event

rizkiarm:
This will add a listener for ``user_change`` and hubot brain ``loaded`` event. Both listener will reload the user data in hubot brain to maintain integrity.

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -195,7 +195,7 @@ class SlackBot extends Adapter
 
     # also wipe out any broken users stored under usernames instead of ids
     for id, user of @robot.brain.data.users
-      if id is user.name then delete @robot.brain.data.users[user.id]
+      if id is user.name then delete @robot.brain.data.users[id]
 
 
 module.exports = SlackBot
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -188,17 +188,14 @@ class SlackBot extends Adapter
   ###
   brain_loaded: () =>
     @robot.logger.info "Brain loaded, reloading all users"
-    self = this
+    
+    # once the brain has loaded, reload all the users from the client
+    for id, user of @client.rtm.dataStore.users
+      @user_change { user: user }
 
-    @client.web.users.list((err,response) ->
-      # once the brain has loaded, reload all the users from the client
-      for id, user of response.members
-        self.user_change { user: user }
-
-      # also wipe out any broken users stored under usernames instead of ids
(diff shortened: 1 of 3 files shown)


DEGoodmanWilson:
Howdy, thanks for this! Looking over it now!

DEGoodmanWilson:
One thing that we will want is test coverage over this new feature, so you know!

johnagan:
I'm pretty sure v3 of the node-sdk stores the users in the dataStore immediately after authenticating and updates it accordingly. It's worth another set of 👀 though.

https://github.com/slackhq/node-slack-sdk/tree/5c8d3bdbcc38a71ca45d372444ce5f79178de557/lib/data-store/message-handlers

Additionally, we only used this to convert IDs <> mentions, which I think we can now handle in the API with `parse` and `link_names`

rizkiarm:
@DEGoodmanWilson: Okay, will do it when I have time to spare

@johnagan: Confirmed, dataStore stores all kinds of information upon authenticating, I'll try to change the implementation accordingly, soon. 

Btw, any other idea on how to provide the other hubot scripts access to important data (users, channels, etc) in the dataStore without using using hubot brain?

DEGoodmanWilson:
I find myself suddenly concerned about the two layers of cacheing involved. I wonder if there is a more general mechanism available that would simply have the brain object pass through queries about Slack objects to dataStore instead of trying to keep the two in sync?

rizkiarm:
It would be great if there is such mechanism. However, as for now I don't have any other ideas to implement this functionality without sacrificing the compatibility issues that might arise with the other scripts (e.g: hubot-brain, etc). Based on my limited observation on several other hubot adapters, it seems that almost all of them use this kind of approach, which might indicates that this is the preferred way to handle this things (?).

DEGoodmanWilson:
Looking at the Campfile adapter (and taking it as canonical), it provides a small number of functions for querying Campfire entities (users, rooms) that simply pass through to the underlying APIs, rather than using the brain as a cacheing layer. We've decided that exposing the `node-slack-sdk` client object is the right thing to do, exposing the existing cache that it provides directly.

Part of the motivation for this is that the brain is meant as a persistent store of data for Hubot scripts, but that the data that you are wanting to store there is not meant in large part to be persisted but merely cached, as most of it is volatile. So, I am sorry to say, I am going to turn down your PR as not being what we consider to be in the spirit of the projects' goals and aims.

PR #346, opened 2016-08-22
Accept room names for robot.messageRoom

mdelagrange:
* [X] I've read and understood the [Contributing guidelines](./CONTRIBUTING.md) and have done my best effort to follow them.
* [X] I've read and agree to the [Code of Conduct](./CODE_OF_CONDUCT.md).
* [X] I've been mindful about doing atomic commits, adding documentation to my changes, not refactoring too much.
* [X] I've a descriptive title and added any useful information for the reviewer. Where appropriate, I've attached a screenshot and/or screencast (gif preferrably).
* [X] I've written tests to cover the new code and functionality included in this PR.
* [X] I've read, agree to, and signed the [Contributor License Agreement (CLA)](https://docs.google.com/a/slack-corp.com/forms/d/1q_w8rlJG_x_xJOoSUMNl7R35rkpA7N6pUkKhfHHMD9c/viewform).

#### PR Summary
Previous to hubot-slack 4.0.0, `robot.messageRoom` would accept room names.  This PR translates room names to room ids, in order to improve compatibility with existing hubot scripts.

#### Test strategy
Write a hubot script that sends a message to a room by name, e.g.:
```coffeescript
module.exports = (robot) ->

  robot.respond /test messageRoom/, (msg) ->
    room = 'general'
    robot.messageRoom room, 'hi there!'
```
This will not work with hubot-slack 4.0.2.

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -73,7 +73,7 @@ class SlackClient
   send: (envelope, message) ->
     message = @format.outgoing(message)
     room = envelope.room
-    if !(room.match /C0[A-Z0-9]{7}/)
+    if !(room.match /[A-Z]/) # slack rooms are always lowercase
       # try to translate room name to room id
       channelForName = @rtm.dataStore.getChannelByName(room)
       if channelForName
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -7,13 +7,13 @@ SLACK_CLIENT_OPTIONS =
 
 
 class SlackClient
-  
+
   constructor: (options) ->
     _.merge SLACK_CLIENT_OPTIONS, options
 
     # RTM is the default communication client
     @rtm = new RtmClient options.token, options
-    
+
     # Web is the fallback for complex messages
     @web = new WebClient options.token, options
 
@@ -36,7 +36,7 @@ class SlackClient
   ###
   on: (name, callback) ->
     @listeners.push(name)
-    
+
     # override message to format text
     if name is "message"
       @rtm.on name, (message) =>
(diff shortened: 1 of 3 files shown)


Mezzle:
:+1:

Fixes: #349

Mezzle:
Actually, personally I'd prefer this with the `#` channel prefix demarking the fact that it's a channel name, or using the `RECOGNIZED_API_TYPE_PREFIXES` to determine it's not an API prefix?

src/client.coffee
```diff
@@ -72,13 +72,19 @@ class SlackClient
   ###
   send: (envelope, message) ->
     message = @format.outgoing(message)
+    room = envelope.room
+    if !(room.match /C0[A-Z0-9]{7}/)
```
Mezzle: Channels begin with `C` not `C0`

see https://github.com/slackhq/node-slack-sdk/blob/master/lib/models/helpers.js#L28

mdelagrange:
> Actually, personally I'd prefer this with the # channel prefix demarking the fact that it's a channel name, or using the RECOGNIZED_API_TYPE_PREFIXES to determine it's not an API prefix?

Actually `getChannelByName(room)` works with or without the hash, so your example should work fine.

mdelagrange:
> Channels begin with C not C0

Since Slack channels never have capital letters, I'm just testing for those now.  That should be a reliable test.

Mezzle:
@mdelagrange I would expect that if I put in 'GENERAL'  that it would push to `#general` ?

mdelagrange:
That wasn't the behavior in hubot-slack < 4.0.0.  Channel names were case sensitive in previous versions.  I'd like to avoid an unnecessary API call, so that's my reason for sticking with case sensitivity.

DEGoodmanWilson:
LGTM

markmcintyre:
@DEGoodmanWilson: I just tested out this fix, which works great; but (and forgive me if I'm misunderstanding the process here) the package.json version number hasn't been updated. This means the latest version on NPM doesn't have this fix included. Is that intentional?

DEGoodmanWilson:
Right, I don't think I released this. Last two weeks have been a bit hectic :( Will look into this ASAP

PR #370, opened 2016-11-05
Renaming a function that has a name collision with the base class

DEGoodmanWilson:
So, it turns out that the `robot` base class has an event handler named `topic`, and the `bot` derived class overrides that into a different method signature, and this breaks things.
https://dev4slack.slack.com/archives/sdk-hubot-slack/p1478282324000005 and _ff._

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -180,7 +180,7 @@ class SlackBot extends Adapter
 
       when 'channel_topic', 'group_topic'
         @robot.logger.debug "#{user.name} set the topic in #{channel.name} to #{topic}"
-        @receive new TopicMessage user, message.setTopic, message.ts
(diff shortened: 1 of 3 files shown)


src/bot.coffee
```diff
@@ -180,7 +180,7 @@ class SlackBot extends Adapter
 
       when 'channel_topic', 'group_topic'
         @robot.logger.debug "#{user.name} set the topic in #{channel.name} to #{topic}"
-        @receive new TopicMessage user, message.topic, message.ts
+        @receive new TopicMessage user, message.setTopic, message.ts
```
aoberoi: @DEGoodmanWilson looks like you might have a search and replace mistake. this line should not change.

PR #371, opened 2016-11-08
And some things that should not have been forgotten were lost.

DEGoodmanWilson:
Stuff that should _not_ have been pushed with the recent documentation update.


aoberoi:
Not that it matters now, but :+1:

PR #376, opened 2016-11-11
Restore the original upgrading docs so they can be added to

DEGoodmanWilson:
Restores the upgrading documentation that got nuked with the recent documentation upgrade.

### DIFF
--- a/docs/_pages/basic_usage.md
+++ b/docs/_pages/basic_usage.md
@@ -4,7 +4,6 @@ title: Basic Usage
 permalink: /basic_usage
 order: 4
 headings:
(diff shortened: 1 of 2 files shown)


aoberoi:
tiny nit, but maybe put upgrading from version 3 above upgrading from version 2? i'm guessing that there's more people who are likely at version 3 already and need to move up one major version, than the number of people who are still at version 2 and need to move up two major versions.

DEGoodmanWilson:
Fair enough, this is basically a copy-and-paste job from the original documentation with very little editing. I'll reverse the order.

Issue #181, opened 2015-04-15
how to get message without url


rb1980:
I just started playing with hubot-slack, so forgive if this is a noob question.
I tried setting up this simple example to test with:
https://github.com/sapg/hubot-scripts/blob/master/shell-command.coffee
I type: host lookup www.domain.com
but I am getting back:
Looking up http://www.domain.com...
It seems that slack is parsing the domain name automatically and I am getting a url back in hubot rather than the original string. Is there a way to get the message "non-url-ified?"

tcrammond:
I'm having a similar problem. A hubot script I have will reply to a command with `https://somedomain.com`, but when it's posted in slack it will be `https://http://somedomain.com`.

Is there a way around this?

jiang-wei:
any plan to fix this bug?

DEGoodmanWilson:
This shouldn't be a problem. `hubot-slack` strips out the link formatting, returning a message text that if you type
`host lookup www.domain.com`
the script should receive message text
`host lookup www.domain.com`
despite whatever link decoration that Slack adds to the message. (with the original formatted message available in the message rawText)

Can you confirm to me that this is still a bug with steps to reproduce?

johnagan:
closing since no response.

rb1980:
sorry for not posting a follow-up.
I haven't updated hubot in a while, so don't know if this was fixed in a more recent version, but a simple workaround I found a while back was this:

`hostname = msg.match[1].replace /http:\/\//, ""`

mister-red:
@rb1980 I know its been a while since you probably thought about this, but I have a more proper workaround, if it works for you

`            /* Get the raw text of the arguments */
            var args_string = msg.message.rawText.trim();

            /* Parse any auto generated http links from slack that were not intentional and just replace it with the full text */
            var httplink = /<(http:\/\/.*)\|(.*)>/i
            var http_match = args_string.match(httplink)
            if (http_match) { args_string  = args_string.replace(http_match[0], http_match[2]) }
`

PR #413, opened 2017-04-19
Return rawText and rawMessage objects

mistydemeo:
* [x] I've read and understood the [Contributing guidelines](./CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](./CODE_OF_CONDUCT.md).
* [x] I've been mindful about doing atomic commits, adding documentation to my changes, not refactoring too much.
* [x] I've a descriptive title and added any useful information for the reviewer. Where appropriate, I've attached a screenshot and/or screencast (gif preferrably).
* [x] I've written tests to cover the new code and functionality included in this PR.
* [x] I've read, agree to, and signed the [Contributor License Agreement (CLA)](https://docs.google.com/a/slack-corp.com/forms/d/1q_w8rlJG_x_xJOoSUMNl7R35rkpA7N6pUkKhfHHMD9c/viewform).

#### PR Summary
Prior to 900402ea254b0fdf5a819fd3073325c68f1692a9, hubot-slack returned a message object which included the raw text from the Slack API along with the raw Slack Message object. This was useful for middleware and scripts which reformatted Slack text in different ways than hubot-slack's own default formatter. This object was a subclass of hubot's `TextMessage` which was otherwise compatible.

For example, a middleware might try to detect URLs that were autoexpanded by the Slack API and restore them to their original format; this is only possible if the raw text is included so the unparsed URLs can be read.

#### Related Issues
#181 also discusses using this property (at the time when it still existed).

#### Test strategy
Added a test to the `SlackBot.message` code.

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -149,7 +149,7 @@ class SlackBot extends Adapter
   Message received from Slack
   ###
   message: (message) =>
-    {text, rawText, user, channel, subtype, topic, bot} = message
+    {text, rawText, returnRawText, user, channel, subtype, topic, bot} = message
 
     return if user && (user.id == @self.id) # Ignore anything we sent, or anything from an unknown user
     return if bot && (bot.id == @self.bot_id) # Ignore anything we sent, or anything from an unknown bot
@@ -174,7 +174,10 @@ class SlackBot extends Adapter
 
       when 'message', 'bot_message'
         @robot.logger.debug "Received message: '#{text}' in channel: #{channel.name}, from: #{user.name}"
-        textMessage = new SlackTextMessage(user, text, rawText, message)
+        if returnRawText
+          textMessage = new SlackTextMessage(user, text, rawText, message)
+        else
+          textMessage = new TextMessage(user, text, message.ts)
         textMessage.thread_ts = message.thread_ts
         @receive textMessage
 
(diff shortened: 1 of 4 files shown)


mistydemeo:
re the @CLAassistant check, my employer has signed a corporate CLA.

stevenjm:
+1 to this.

We are now using this patch in production to deal with FQDNs that Slack incorrectly interprets as HTTP links, and it would be good to get it merged so we can drop our custom-built hubot-slack package. Please let me know if there's anything I can do to help progress this.

aoberoi:
thanks for the contribution @mistydemeo!

i'll override the CLAbot when we merge, since I have the record of GitHub's signed contributor agreement.

I think having `rawText` and `rawMessage` available on the messages is a great idea, and thank you for implementing it. My one reservation is regarding `rawMessage`. On very busy slack teams, this could cause an order magnitude more objects to be created (and garbage collected) so there would be a performance impact. What I'd like to see is a way of turning the option on when initializing the adapter, and therefore making this opt-in. Would you be willing to work on adding that?

mistydemeo:
Since this was the old behaviour, and the change doesn't seem to have been documented, how about making it opt-out instead of opt-in? I can also say that we're running this on a busy Slack team, and it hasn't been an issue for us.

aoberoi:
@mistydemeo okay, i like your rationale for opt-out. happy to merge that if you can make the change.

mistydemeo:
Added a commit which allows disabling this behaviour.

mistydemeo:
Thanks!

PR #422, opened 2017-06-27
Accept DM'ed commands with or without bot name prefix

patcon:
See: https://github.com/hubotio/hubot/issues/1356

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -216,8 +216,13 @@ class SlackBot extends Adapter
 
 
     # Direct messages
-    if channel.id[0] is 'D'
-      text = "#{@robot.name} #{text}"     # If this is a DM, pretend it was addressed to us
+    isDM = channel.id[0] is 'D'
+    if isDM
+      startOfText = if text.indexOf('@') == 0 then 1 else 0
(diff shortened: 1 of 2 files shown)


aoberoi:
@patcon thanks for the contribution and my apologies for the late feedback!

i think this is a great idea. in fact, #291 seems to implement the same behavior. i also think the environment variable approach (discussed in hubotio/hubot#1356) in addition would be an even greater improvement.

if you can contribute some test code, i'd be happy to merge this. if you'd like help with getting that done, let me know.

aoberoi:
ping, still open to this contribution!

patcon:
gr... i gitted wrong

patcon:
okie doke fixed

aoberoi:
@shanedewael would merging this make your work in #465 more complicated? one opportunity to simplify i see is that instead of matching on the first letter of the channel ID, since we have the entire `conversations.info` object, we should be able to use `channel.is_im || channel.is_mpim` instead.

shanedewael:
@aoberoi This should definitely be able to work with #465 - merge ahead 🎉

PR #463, opened 2018-02-03
Fix "ERROR Can't fetch users"

leschekfm:
###  Summary

With commit 66b31784d72dd3352ecd78106b8fd492a9193428 a paginated loadUsers() function was introduced in SlackClient which changed the param that is provided to SlackBot.loadUsers()

Apparently the stubs in the tests weren't adjusted accordingly and also the receiving function wasn't updated. This results in a error that is logged everytime hubot starts `ERROR Can't fetch users`


src/bot.coffee
```diff
@@ -222,8 +222,9 @@ class SlackBot extends Adapter
     user.room = item.channel
     @receive new ReactionMessage(type, user, reaction, item_user, item, event_ts)
 
+  # Callback for SlackClient.loadUsers()
   loadUsers: (err, res) =>
-    if err || !res.ok
+    if err || !res.members.length
```
Roach: If the user is cnnecting via `rtm.connect`, there's a valid case where `members` will be empty

aoberoi:
good catch!

PR #464, opened 2018-02-16
Add display_name to user object

aahill50:
###  Summary

Slack users now have a `profile` object which keeps track of the name they are displayed as in slack. This PR adds that field to the user object

### Requirements

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -237,6 +237,7 @@ class SlackBot extends Adapter
     newUser =
       id: user.id
       name: user.name
+      display_name: user.profile.display_name
       real_name: user.real_name
       slack: {}
     newUser.email_address = user.profile.email if user.profile and user.profile.email
--- a/test/bot.coffee
+++ b/test/bot.coffee
@@ -306,16 +306,6 @@ describe 'Users data', ->
     should.equal user.email_address, @stubs.user.profile.email
     should.equal user.slack.misc, @stubs.user.misc
 
-  it 'Should add a user data (user with no profile)', ->
-    @slackbot.userChange(@stubs.usernoprofile)
-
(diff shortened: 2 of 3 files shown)


aahill50:
This conforms with the changes laid out [here](https://api.slack.com/changelog/2017-09-the-one-about-usernames#changing)

Roach:
Thanks @aahill50! Travis tests are failing because `display_name` needs to be added to the  user stubs: https://github.com/slackapi/hubot-slack/blob/master/test/stubs.coffee

aoberoi:
i'm sorry to be the party pooper, but i feel like this is not a change we should land.

ideally, you don't want to depend on `display_name` for any programmatic use, similar to `username`, as discussed in the changelog entry linked above. raising it into a property of the user only promotes the pattern of use that we want to discourage. i can't think of a time you should use the display name in a script, even when you are formatting a user mention, all you should require is the user ID. after #465, you'll even have access to user IDs for mentioned users, so it should be easy to format a mention in a text string.

i do hope you understand this point of view, and i'm sorry we cannot take this contribution. a big thanks for updating the tests, that was something most contributors ignore. if you're still interested in working on this project, there's a few more issues labelled `good first issue` that you might want to check out.

src/bot.coffee
```diff
@@ -237,6 +237,7 @@ class SlackBot extends Adapter
     newUser =
       id: user.id
       name: user.name
+      display_name: user.profile.display_name
```
aoberoi: sorry, expanding the user API to include `display_name` is not something we are open to doing at this time.

aahill50:
@aoberoi No problem! I actually came to that realization shortly after opening this PR and updated our usage accordingly, I just figured I would leave this PR open in case it was useful for anyone else, but I can understand if the usage is not supported. Thanks for the update

Issue #385, opened 2016-12-29
remove dependency on SlackClient.rtm.dataStore (MemoryStore)


aoberoi:
## Description

While the SlackClient instance has access to the underlying `@slack/client`'s `RtmClient.dataStore`, using it to store data that is critical to the operation of this adapter is not wise. The `dataStore` abstraction isn't designed well to handle synchronization and persistence use cases. It isn't guaranteed to contain all team information (since the introduction of Enterprise Grid). Fortunately, hubot provides `robot.brain` as a useful abstraction for these purposes. We should cease to depend on the dataStore for internal operations and instead use `robot.brain` where ever it makes sense.

In v4.3.0, we began to put user data inside the robot brain. We need to take inventory of all other places where data is being fetched from the dataStore for internal operations, identify alternatives, and implement those alternatives.

Subtasks:

- [ ] announce the deprecation of the `adapter.client.rtm.dataStore` APIs.
- [ ] identify usages of dataStore originated data. determine alternative implementations. implement these changes.
- [ ] write a migration guide to help users onboard onto alternatives when they find that the API is deprecated.

Areas to be refactored:
1. `SlackFormatter`: The class can be left as-is and even instantiated and referenced from `SlackClient`. However, it will never be called. It's functionality will be reimplemented as according to #456

Note: in a future version, we should deprecate the use of the `RtmClient` directly, possibly by making `SlackClient.rtm` a private property.

* [x] I've read and understood the [Contributing guidelines](./CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](./CODE_OF_CONDUCT.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

sebastiandero:
I would suggest a simple `robot.logger.warn "You are using a deprecated access to the user list, this feature will be removed in #{version} "`
At some point maintaining both access points might lead to inconsistent data or performance.

aoberoi:
@sebastiandero i'd rather not wrap the entire DataStore API in a method that logs before forwarding the function call into the real DataStore object. I think announcing that it is deprecated and making sure its not removed before a semver:major change should be enough. when a major change is released, we should have a migrating guide that will help users correctly update their code/scripts.

the adapter has been supporting the hubot brain since v4.3.0. i'll update the issue for reflect the remaining work.

aoberoi:
another dependency to eliminate:

[...]

PR #465, opened 2018-03-21
Decouple RTM Datastore

shanedewael:
###  Summary

Removes RTM datastore from Hubot (addresses #385). This necessitates adding calls to fetch user and channel info using Slack client. Deprecates Formatter in favor of SlackTextMessage and added test cases for SlackTextMessage. Adds support for reaction events in shared channels.


.gitignore
```diff
@@ -3,3 +3,4 @@ package-lock.json
 .idea
 coverage/
 *.log
+.DS_Store
```
aoberoi: pet peeve of mine, but this should really be in your global `gitignore` since its a platform dependent file

src/bot.coffee
```diff
@@ -1,6 +1,8 @@
-{ Adapter, TextMessage, EnterMessage, LeaveMessage, TopicMessage, CatchAllMessage, Robot } = require.main.require 'hubot'
+{ Adapter, TextMessage, EnterMessage, LeaveMessage, TopicMessage, CatchAllMessage } = require.main.require 'hubot'
 { SlackTextMessage, ReactionMessage } = require './message';
+{ Robot } = require './extensions';
```
aoberoi: the `extensions.coffee` file doesn't have any exports, so this seems incorrect ( `Robot` will just be set to `undefined`). `Robot` isn't really used directly anywhere in this file, so i guess its safe to delete this line. the only side effect was that requiring this file extended the `Robot` class with the `react()` prototype method. We get that from line 2 of `slack.coffee`. If load order is an issue (i'm honestly not sure) then maybe that should be moved to line 1 instead.

src/bot.coffee
```diff
@@ -188,14 +207,28 @@ class SlackBot extends Adapter
...
+            @receive message
+          )
+          .catch((error) =>
+            @robot.logger.error "Error constructing SlackTextMessage"
+          )
+        else
+          @receive new CatchAllMessage {user}
+
     else if event.type is 'reaction_added' or event.type is 'reaction_removed'
+      return unless event.user && event.item_user
```
aoberoi: how could this ever happen?

src/bot.coffee
```diff
@@ -188,14 +207,28 @@ class SlackBot extends Adapter
...
+
+      # If the reaction is from a stranger or on a stranger's message, then the event will be ignored
+      # Strangers have no shared channel in common, external members have shared channel membership
+      # External member reactions should function normally
+      return if event.item_user.is_stranger || event.user.is_stranger
```
aoberoi: this definitely reflects what [the docs](https://api.slack.com/shared-channels#users_may_seem_strange) say, but i think this is worth double checking in an actual workspace with shared channels. from my memory trying it out, this doesn't seem right.

src/client.coffee
```diff
@@ -1,7 +1,6 @@
 {RtmClient, WebClient} = require '@slack/client'
 SlackFormatter = require './formatter'
 _ = require 'lodash'
-Promise = require 'bluebird'
```
aoberoi: you need this line since Promise is used in this file.

src/message.coffee
```diff
@@ -33,22 +35,25 @@ class SlackTextMessage extends TextMessage
   # text       - The parsed message text
   # rawText    - The unparsed message text
   # rawMessage - The Slack Message object
-  constructor: (@user, text, rawText, @rawMessage, channel, robot_name) ->
+  constructor: (@user, text, rawText, @rawMessage, channel, robot_name, client) ->
```
aoberoi: rather than saving a reference to the client and then deleting it later (good call: don't want reference cycles), how about passing the client into the `buildText()` method? it could be `buildText(client, cb)`.

src/message.coffee
```diff
@@ -33,22 +35,25 @@ class SlackTextMessage extends TextMessage
...
     @_channel = channel
     @_robot_name = robot_name
 
+    # Slack client is needed to get user and conversation info
+    @_client = client
+
     # public instance property initialization
     @rawText = if rawText? then rawText else @rawMessage.text
-    @text = if text? then text else @buildText()
+    @text = if text? then text else undefined # else @buildText()
```
aoberoi: delete the part that's commented out.

src/message.coffee
```diff
@@ -57,45 +62,96 @@ class SlackTextMessage extends TextMessage
...
+    return client.web.conversations.info(id).then((channel) -> "\##{channel.name}")
 
-    @text = text
+  ###*
+  # Factory method to construct SlackTextMessage
+  ###
+  @makeSlackTextMessage: (@user, text, rawText, @rawMessage, channel, robot_name, client, cb) ->
+    message = new SlackTextMessage(@user, text, rawText, @rawMessage, channel, robot_name, client)
 
+    if not message.text? then message.buildText(() =>
```
aoberoi: you can't use a JavaScript arrow function (`() => { }`) if you want this to work in older versions of node. you can, however, use CoffeeScript arrow functions because the coffeescript compiler will turn it into a regular `function() { }` expression.

src/message.coffee
```diff
@@ -57,45 +62,96 @@ class SlackTextMessage extends TextMessage
...
+      message._client = undefined
+      cb(message)
+    ) else 
+      message._client = undefined
+      cb(message)
```
aoberoi: one piece of advice - put this call to `cb()` in a `setImmediate()` so that this function is always deterministically async. the way its written now, depending on the value of `text`, the callback is called either after the function returns or before the function returns, and that can lead to non-deterministic behavior for the caller.

src/message.coffee
```diff
@@ -57,45 +62,96 @@ class SlackTextMessage extends TextMessage
...
+  ###*
+  # Replace links inside of text
+  ###
+  replaceLinks: (text) ->
+    parts = []
+    cursor = 0
+    regex = SlackTextMessage.MESSAGE_REGEX
+    regex.lastIndex = 0
+
+    while (res = regex.exec(text))
```
aoberoi: what happened to the [string-replace-async](https://www.npmjs.com/package/string-replace-async) module? this loop looks a little complex.

src/message.coffee
```diff
@@ -57,45 +62,96 @@ class SlackTextMessage extends TextMessage
...
-          if label then return "@#{label}"
-          # TODO
-          # user = @dataStore.getUserById link
-          # if user
-          #   return "@#{user.name}"
-          return m
+          if label
+            parts.push(text.slice(cursor, res.index), "@#{label}")
+          else
+            parts.push(text.slice(cursor, res.index), @replaceUser.apply(null, [link, @_client]))
```
aoberoi: why was `apply` necessary?

src/message.coffee
```diff
@@ -57,45 +62,96 @@ class SlackTextMessage extends TextMessage
...
+    )
 
-    text = text.replace /&lt;/g, '<'
-    text = text.replace /&gt;/g, '>'
-    text = text.replace /&amp;/g, '&'
+  ###*
+  # Returns name of user with id
+  ###
+  replaceUser: (id, client) ->
+    return client.web.users.info(id).then((user) -> "@#{user.name}")
```
aoberoi: you're throwing away the rest of the user object (besides the name property). i thought we wanted to tack this onto the message as a property somewhere so that the receiver had the data about who was mentioned?

src/message.coffee
```diff
@@ -57,45 +62,96 @@ class SlackTextMessage extends TextMessage
...
 
-    if @_channel?.is_im
-      text = "#{@_robot_name} #{text}"     # If this is a DM, pretend it was addressed to us
+  ###*
+  # Returns name of channel with id
+  ###
+  replaceChannel: (id, client) ->
+    return client.web.conversations.info(id).then((channel) -> "\##{channel.name}")
```
aoberoi: ditto to above.

aoberoi:
okay i know there's a lot of comments. i didn't make it to the tests because i figured a lot might change once you address those comments. overall i think its a good first stab! it seems like there's parts you copied in from previous commits for the sake of not changing tests, but the underlying issue was that the tests were bad. not your fault, but lets talk about why certain changes were made before you copy code from previous commits again.

.gitignore
Roach: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

```
git config --global core.excludesfile ~/.gitignore_global
```

src/bot.coffee
```diff
@@ -162,14 +155,17 @@ class SlackBot extends Adapter
   # @private
   ###
   eventHandler: (event) =>
-    {text, rawText, user, channel, subtype, topic, bot, item} = event
+    {user, channel, bot} = event
 
     # Ignore anything we sent
     # NOTE: coupled to getting `rtm.start` data
-    return if (user && (user.id is @self.id)) || (bot && (bot.id is @self.bot_id))
+    return if (user && (user.id is @self.id || user.id is @self.bot_id)) || (bot && (bot.id is @self.bot_id))
```
aoberoi: this is interesting, can `user.id === @self.bot_id`? when would that happen?

src/bot.coffee
```diff
@@ -181,14 +177,14 @@ class SlackBot extends Adapter
...
-      switch subtype
+      switch event.subtype
         when 'bot_message'
           @robot.logger.debug "Received message in channel: #{channel.name || channel.id}, from: #{user.name}"
 
           # prefer user over bot.
           # if both are set in the slack event, it represents an app or integration sending a message on behalf of a
           # user, so the user is the more appropriate value.
-          SlackTextMessage.makeSlackTextMessage(user || bot, undefined, undefined, event, channel, @robot.name, @client, (message) =>
+          SlackTextMessage.makeSlackTextMessage(user, undefined, undefined, event, channel, @robot.name, @client, (message) =>
```
aoberoi: when this is a message with subtype `bot_message` (the case we are in), the user is only set when the app or integration is sending a message on behalf of the user (like when using `as_user: true`), otherwise it will be undefined, and in that case the `bot` should be the fallback, right?

src/message.coffee
```diff
@@ -131,26 +134,43 @@ class SlackTextMessage extends TextMessage
...
-    return client.web.users.info(id).then((user) -> "@#{user.name}")
+  replaceUser: (client, id, mentions) ->
+    client.web.users.info(id).then((user) ->
+      if user
+        mention = new SlackMention(user.id, 'user', user)
+        mentions.push(mention)
+        return "@#{user.name}"
+      else return link
+    )
+    .catch((error) =>
```
aoberoi: if this error happens, what impact will there be on the result of `replaceLinks()`?

src/message.coffee
```diff
@@ -131,26 +134,43 @@ class SlackTextMessage extends TextMessage
...
-    return client.web.conversations.info(id).then((channel) -> "\##{channel.name}")
+  replaceConversation: (client, id, mentions) ->
+    client.web.conversations.info(id).then((conversation) ->
+      if conversation
+        mention = new SlackMention(conversation.id, 'conversation', conversation)
+        mentions.push(mention)
+        return "\##{conversation.name}"
+      else return link
+    )
+    .catch((error) =>
```
aoberoi: ditto to above

src/bot.coffee
shanedewael: With reaction events the user.id is the bot id

src/bot.coffee
shanedewael: line 176 is setting the fallback to be the `bot` if the user is undefined, regardless of the `as_user` param

src/message.coffee
shanedewael: Good call. I modified the catch to return back the original link (`<@Uxxx>` or `<#Cxxx>`) which I would assume is desired behavior.

src/bot.coffee
aoberoi: maybe we need to back up a second and think about what that means.

the value of `bot` is jut a `bot_id` (`Bxxxx`), right? the value of `user` is the response of `users.info` (a full JSON object), right?

in either case, isn't doing the substitution on every message unnecessary? don't we only want to look into the bot object if we know the message is of subtype `bot_message`?

src/message.coffee
```diff
@@ -0,0 +1,192 @@
...
+    super @user
+    @type = @type.replace('reaction_', '')
+
+class PresenceMessage extends Message
+  # Represents a message generated by a presence change event
+  #
+  # users      - Array of users that changed their status
+  # presence   - Status is either 'active' or 'away'
+  constructor: (@users, @presence) ->
+    super @users
```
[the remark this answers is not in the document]
aoberoi: or this is a terrible idea. i didn't think much about it.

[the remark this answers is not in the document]
Roach: Would that need to be either a placeholder string (`no-room`) or the message ID? 🤔

src/bot.coffee
```diff
@@ -237,6 +237,8 @@ class SlackBot extends Adapter
         user = @robot.brain.data.users[id]
         if user then users.push user
 
+        console.log(new PresenceMessage(users, event.presence))
```
Roach: Intentional log statement?

shanedewael: No, I'm just terrible

Roach: (づ￣ ³￣)づ

PR #470, opened 2018-04-03
Fix bug using robot's name for prepending

shanedewael:
###  Summary

Fixes an incorrectly merged community contribution that wasn't using the correct bot name.

### DIFF
--- a/src/message.coffee
+++ b/src/message.coffee
@@ -79,7 +79,7 @@ class SlackTextMessage extends TextMessage
 
       if @_channel?.is_im
         startOfText = if text.indexOf('@') == 0 then 1 else 0
-        robotIsNamed = text.indexOf(@robot.name) == startOfText || text.indexOf(@robot.alias) == startOfText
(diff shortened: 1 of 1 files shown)


src/message.coffee
```diff
@@ -79,7 +79,7 @@ class SlackTextMessage extends TextMessage
 
       if @_channel?.is_im
         startOfText = if text.indexOf('@') == 0 then 1 else 0
-        robotIsNamed = text.indexOf(@robot.name) == startOfText || text.indexOf(@robot.alias) == startOfText
+        robotIsNamed = text.indexOf(@_robot_name) == startOfText || text.indexOf(client.robot.alias) == startOfText
```
aoberoi: it seems a little contradictory to me that we're both passing in a property of the robot instance (`name`) in the constructor to avoid accessing the `client.robot` instance directly, and also using that interface to access another property (`alias`). we should decide if we want to access the `client.robot` value at all, and the be consistent.

i'd suggest that we don't access it, because that kind of tight coupling makes testing the `SlackTextMessage` object more difficult.

PR #471, opened 2018-04-04
Add flag to disable loading all users on hubot startup

### LINKED ISSUE #467: Flag to disable loading all users on hubot startup (saves space in backing store)
### Description
I understand that #381 was done with good intent but the Slack team I run this bot in has more than 10,000 users and putting them all in the brain crosses the limit of any free-tier Redis backend on Heroku because it takes up more than 30MB.  I'm not willing to pay for more memory until I can prove the bot's usefulness as proof of concept and the MongoDB brain I tried didn't work at all.  For now I've downgraded from 4.4.0 to 4.2.2, before that PR was merged, so that I can have a persistent brain for other non-user things.

Obviously this is a needed feature for some, since it was re-implemented at all, so it can't just be turned off, but maybe saving user data could be made either **toggleable** or **filterable**.  The toggleability might be the simpler of the two:  If the user sets an env var like `NO_LOAD_USERS` the code path that loads the users to the brain is not run.  As a more flexible solution, since in my specific case I know that only ~100 of the ~10,000 users in the team will use this bot at first, it might be worth being able to turn on loading the users to the brain in such a way that the users loaded are filterable by some kind of grouping or categorisation, e.g. they all belong to a named Slack channel.

If we can agree on which of these solutions would be preferable, or someone else has an even better idea, maybe I could try and do a pull request for it because it would be nice to use v4.4.0 without my Redis exploding. :stuck_out_tongue_closed_eyes: 

- [X] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [X] discussion

### Requirements
* [X] I've read and understood the [Contributing guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [X] I've searched for any related issues and avoided creating a duplicate issue.

sinisterstuf:
###  Summary

For a team with a large number of users, auto-loading all users uses a lot of space in the Hubot brain.  This commit adds an environment flag: `NO_LOAD_ALL_USERS`.  When set, the `loadUsers` event will NOT trigger `userChange` events for all users, instead it will be skipped and users will only be loaded on-demand, when the bot hears them say something.

Fixes #467

### Requirements


sinisterstuf:
@aoberoi here you go, I basically took the error checking code above it as an example, and made it exit early. Got a better idea? :stuck_out_tongue: 

Need me to add any test case or mention this env var in the README or something? The user loading tests that existed so far passed fine "on my machine".

sinisterstuf:
To be honest I haven't actually tested that this solves my problem, but if there's a doubt about that I can try it out tomorrow.

aoberoi:
whoops! i just commenting in your issue before seeing the PR.

feedback:
*  can we keep all the environment variable parsing in `slack.coffee`? passing values into the `SlackBot` object instead of it reading the env helps with testability.
*  this will bail before writing to the brain, but we can save some network performance cost if we never even make the `users.list` API call, right?
*  we'll need some tests before we merge this in.

sinisterstuf:
> just commented in your issue before seeing the PR

I think your comment and the PR creation happened within seconds of each other haha.

> see feedback in comment

Thanks for the feedback, I'll give it another shot later today, shouldn't be too hard.

sinisterstuf:
Moved the env var checking to where the others are. Weirdly, before changing any code, I re-ran the tests and now 2 tests related to loading users are failing, not sure what's up with that, but I guess it'll take a bit more time to investigate...

As for this comment:

> never even make the users.list API call

You mean [this one](https://github.com/sinisterstuf/hubot-slack/blob/no-load-all-users/src/client.coffee#L153)? That would probably work, but then that case would have to have a reasonable no-value return value that won't break things expecting it to return a list of users. I also don't know what else uses it, if it's only used by this, then sure. Maybe I don't have a good overview of the code as a whole yet, so it's not clear for me what uses this and when.

To be honest I actually I have no issue with it loading all the users into memory or something, it just mustn't persist them in the Hubot Brain between deployments :laughing: However, if that's the only thing this is triggered on then I agree it would be best to move the logic there.

aoberoi:
the environment variable and option for `SlackBot` initialization look good. my one suggestion for improvement is to allow the default behavior (if the instance is only initialized with a token) to be consistent with the current behavior. that would make turning the `loadAllUsers` option into something like `disableUserSync` which the `NO_LOAD_ALL_USERS` env var would set to true.

we've landed some newer code on master, so before we can merge this, you'll need to update this branch.

by the "not even make the users.list API call" comment, i'm referring to wrapping any calls to `@client.loadUsers` in a conditional check. i believe there's only two places that need to be touched:

1.  https://github.com/slackapi/hubot-slack/blob/e223126b93a70730fa9e6bbd47c20a240762b232/src/bot.coffee#L47

2.  https://github.com/slackapi/hubot-slack/blob/e223126b93a70730fa9e6bbd47c20a240762b232/src/bot.coffee#L52

i think the easiest solution would be to wrap the line in (1) above with something like this:

```coffeescript
unless @client.disableUserSync
  @client.loadUsers @usersLoaded
else
  @isLoaded = true
```

sinisterstuf:
Thanks for the advice, I'll get to it. I've been struggling to come up for a reasonable name for this and `disableUserSync` is definitely better than the juggling of booleans so far...

sinisterstuf:
Well, at least now the existing tests are passing again... seems it's really better this way. Now just got to solve the coverage drop...

sinisterstuf:
OK, so I thought of 3 test cases that should cover this:

- syncs users normally
- syncs no users when disabled
- will still load a user on demand

Started on that here: 40e04caab12e558f4a851473ba675677a666829f

I found some code where `@slackbot.run` is called in the tests, and thought that's relevant because `run` is where that `unless @client.disableUserSync` logic was added...

Unfortunately I'm stuck right at the first one because it is not clear for me _at all_ where the result of this sync should be observable in the tests. First I assumed it would be `client.dataStore.users` but it seems that's _always_ 4, even if you don't call run. Then I thought maybe it's worth checking the brain, but it seems that's always empty unless you manually load a user with updateUserInBrain as in [this test](https://github.com/slackapi/hubot-slack/blob/master/test/bot.coffee#L348-L356), but that's not the point of this test, I expect that to happen when users are [mass loaded here](https://github.com/slackapi/hubot-slack/blob/master/src/bot.coffee#L254) because of [the call from here](https://github.com/slackapi/hubot-slack/blob/master/src/bot.coffee#L47) in the run function. But apparently not... am I missing a step?

After some poking around I also found `client.web.users` which returns a promise which resolves to... I don't know because the test runner says `Error: invalid_auth`. I could be going on completely the wrong track here so I thought I'd ask because it's probably not that hard if you know what you're looking for and I'm not sure what I'm looking at here.

src/bot.coffee
```diff
@@ -44,15 +44,18 @@ class SlackBot extends Adapter
     # TODO: check this value when connection finishes (even if its a reconnection)
     # TODO: build a map of enterprise users and local users
     @needsUserListSync = true
-    @client.loadUsers @usersLoaded
+    unless @client.disableUserSync
```
aoberoi: does the `@client` object actually have a property called `disableUserSync`? i know its passed in as `options` in the constructor of this object, which means it should be in `@options.disableUserSync`, right?

aoberoi:
i totally understand why this is confusing. the tests don't have a very intuitive design, and i've sort of just been trying to embrace the madness myself.

i agree with the test cases you've listed. that is all the verification we need. i have a recommendation for the first two of them. the technique i propose would be to inspect the stub-injected `_listCount` variable, similar to how its done here:
https://github.com/slackapi/hubot-slack/blob/7bba3e912f426937c61e101cefee0978d82021bb/test/client.coffee#L157-L161

the idea is if synchronization is disabled, then `loadUsers` shouldn't be called. the number of times `loadUsers` is called is instrumented with `_listCount` in the stubs, therefore it should be zero when its disabled and greater than zero when its not disabled.

how do we actually toggle between the two behaviors? well it would have been nice if the `SlackBot` instance wasn't constructed for us in stubs before every test, but we can still "trick" the test into working by setting `@slackbot.options.disableUserSync = true` before `run()` is called, _i think_.

i hope this helps us get moving. i'll keep thinking about how the third case could be verified. in the meantime, would you like to make your branch collaborative?  there's a control you should see on the right column beside this conversation in the PR to allow maintainers to push to the branch. if you decide to do that, i might be able to help get this across the finish line.

src/bot.coffee
sinisterstuf: Yeah, you're right, that's a copy-paste error while I was half asleep from reading the project's test files :laughing: it's supposed to be `@options`

shanedewael:
Hey @sinisterstuf, a few of the PRs I've been working on (#489 and #491) are being merged into master, which I have a feeling will make this change pretty difficult to merge. If you want, I can build on top of your changes and get your commits in a mergeable state. I'd love to see this ship in the next version or two -- I think it'll make a lot of developer's lives a little bit easier 😄

sinisterstuf:
@shanedewael I'd really appreciate that.

Also where I'm stuck with this right now is I can't seem to get a working test case for it.

shanedewael:
@sinisterstuf I was able to work through the testing problems and merged it with the most current master. With #492, I also added you as a contributor. Thanks for your work on this 🎉

Issue #427, opened 2017-07-14
Bot user object missing the is_bot property


Klionsk:
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/hubot-slack/blob/master/CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://github.com/slackapi/hubot-slack/blob/master/CODE_OF_CONDUCT.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description
Bot user objects are missing the `is_bot` property.

#### Reproducible in:
hubot-slack: 4.3.4
hubot: 2.19.0

#### Steps to reproduce:

1. Add a hubot script to capture user objects from all messages
```
    robot.hear /.*/i, (res) ->
        console.log(JSON.stringify res.message.user)
```
2. Put hubot in a channel with another bot
3. Compare the output for messages from human users and bot users

### Expected result:
- Human user objects have `is_bot: false`
- Bot user objects have `is_bot: true`
- `res.message.user.is_bot` is true if a message came from a bot, false if it came from a human

### Actual result:
- Human user objects have `is_bot: false`
- Bot user objects do not have the `is_bot` property.
- `res.message.user.is_bot` always returns false

#### Sample output
```
# Human user
{  
   "user":{  
      "id":"U5D9X9TQR",
      "team_id":"T0CCQK2LA",
      "name":"katie",
      "deleted":false,
      "color":"9e3997",
      "real_name":"Katie",
      "tz":"America/Los_Angeles",
      "tz_label":"Pacific Daylight Time",
      "tz_offset":-25200,
      "profile":{  
         "avatar_hash":"gcdd8e1cb321",
         "first_name":"Katie",
         "last_name":"removed",
         "real_name":"Katie",
         "email":"removed",
         "image_24":"removed",
         "fields":null
      },
      "is_admin":false,
      "is_owner":false,
      "is_primary_owner":false,
      "is_restricted":false,
      "is_ultra_restricted":false,
      "is_bot":false,
      "updated":1499465955,
      "presence":"active"
   },
   "text":"test",
   "id":"1500059368.182916",
   "done":false,
   "room":"G5KJ9CLR3"
}
```
```
# Bot user
{  
   "user":{  
      "id":"B24KXQLQ5",
      "deleted":false,
      "name":"hubot",
      "updated":1472068471,
      "app_id":"A0F7XDU93",
      "icons":{  
[...]

Issue #414, opened 2017-04-24
msg.message.user is now returning Slack User objects, not Hubot User objects


mistydemeo:
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/hubot-slack/blob/master/CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://github.com/slackapi/hubot-slack/blob/master/CODE_OF_CONDUCT.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description
Somewhere between 3.4.0 and 4.3.3, hubot-slack changed its behaviour for returning User objects. In 3.4.0 and earlier versions, the `msg.message.user`/`msg.envelope.user` object is an instance of Hubot's own User object, as stored in the brain (and potentially decorated by other scripts). In the current version, `@slack/client`'s User objects are being returned instead. Since these have different properties with different names, this has the potential to break scripts. I didn't see this documented in the [upgrade guide](https://slackapi.github.io/hubot-slack/upgrading#upgrading-from-version-3-or-earlier), so I thought this might have been an unintentional change.

#### Reproducible in:
{project_name} version: 4.3.3
OS version(s): Any

#### Steps to reproduce:

1. Access the `msg.message.user` object in a Slack script.

### Expected result:
User should be a Hubot User object.

### Actual result:
User is a Slack User object.

mistydemeo:
I suspect this may have changed in 744359dbb92af9b9ac5053b1577ed44302c36cb6 (in release 4.0.0 and later, along with the unreleased 3.5.0). Prior to this commit, the user object is processed into a full hubot user here: https://github.com/slackapi/hubot-slack/commit/744359dbb92af9b9ac5053b1577ed44302c36cb6#diff-44aca805fd3e634b2ed4f6bc2ccb0749L146

```js
    # Process the user into a full hubot user
    user = @robot.brain.userForId msg.user
    user.room = channel.name
```

In 744359dbb92af9b9ac5053b1577ed44302c36cb6, which updated the code for `@slack/client` v3, that code was dropped and the user lookup was moved elsewhere - and turned into a `@client.dataStore.getUserById` call, which returns a Slack User object. Since there was no mention of this in the commit message, and no comment to this effect, I'm pretty sure this was not intentional.

cc @johnagan, who authored the above commit

mistydemeo:
Just checking in - it'd be very helpful to get a response to this.

[...]

PR #472, opened 2018-04-05
Return Hubot Users

shanedewael:
###  Summary

Do some work to standardize the structure of a user inside of the adapter. Modify `loadUsers()` to fetch a Slack user from a bot id. Addresses #414 and #427.

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -204,7 +204,7 @@ class SlackBot extends Adapter
           # prefer user over bot.
           # if both are set in the slack event, it represents an app or integration sending a message on behalf of a
           # user, so the user is the more appropriate value.
-          SlackTextMessage.makeSlackTextMessage(user, undefined, undefined, event, channel, @robot.name, @client, (message) =>
+          SlackTextMessage.makeSlackTextMessage(user, undefined, undefined, event, channel, @robot.name, @robot.alias, @client, (message) =>
             @receive message
           )
         # NOTE: channel_join should be replaced with a member_joined_channel event
@@ -221,7 +221,7 @@ class SlackBot extends Adapter
         when undefined
           @robot.logger.debug "Received message in channel: #{channel.name || channel.id}, from: #{user.name}"
           
-          SlackTextMessage.makeSlackTextMessage(user, undefined, undefined, event, channel, @robot.name, @client, (message) =>
+          SlackTextMessage.makeSlackTextMessage(user, undefined, undefined, event, channel, @robot.name, @robot.alias, @client, (message) =>
             @receive message
           )
         # NOTE: if we want to expose all remaining subtypes not covered above as a generic message implement an else
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -62,7 +62,6 @@ class SlackBot extends Adapter
   # @public
   ###
   send: (envelope, messages...) ->
-    console.log("ENVELOPE #{JSON.stringify(envelope)}")
     for message in messages
       # NOTE: perhaps do envelope manipulation here instead of in the client (separation of concerns)
       @client.send(envelope, message) unless message is ''
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -188,13 +188,14 @@ class SlackBot extends Adapter
     # slack.is_bot {Boolean}:   Flag indicating whether user is a bot
     # name {String}:            Slack username
     # real_name {String}:       Name of Slack user or bot
+    # room {String}:            Channel id of event (will be empty string if no channel)
     ###
     user = if user? then @robot.brain.userForId user.id, user else {}
 
     # Send to Hubot based on message type
     if event.type is 'message'
 
-      user.room = channel?.id
+      user.room = if channel? then channel.id else ''
 
       switch event.subtype
         when 'bot_message'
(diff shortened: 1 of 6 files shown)


src/bot.coffee
```diff
@@ -171,30 +171,38 @@ class SlackBot extends Adapter
...
 
     # Ignore anything we sent
     # NOTE: coupled to getting `rtm.start` data
-    return if (user && (user.id is @self.id || user.id is @self.bot_id)) || (bot && (bot.id is @self.bot_id))
+    return if user && (user?.id is @self.id)
+
+    
+    ###*
+    # Hubot user object in Brain.
+    # User can represent a Slack user OR bot
```
aoberoi: nitpicky, but i think you should say "Slack human user or bot user", and also make it clear that the value is an instance of Hubot's `User` class

src/bot.coffee
```diff
@@ -171,30 +171,38 @@ class SlackBot extends Adapter
...
+    ###
+    user = if user?.id? then @robot.brain.userForId user.id, user else user # Checks if user is just id for presence_change
 
     # Send to Hubot based on message type
     if event.type is 'message'
-      
-      # NOTE: use robot.brain.userForId(id, options) to initialize the user object
-      # think about whether this is true for bots and if we want to be storing bots in the same brain namespace as users
-      # Hubot expects this format for TextMessage Listener
+      # If bot user not found in users.info from client
```
aoberoi: i don't understand this comment. would it be accurate to say "If no user in message, set a default value of an empty object"? and if so, isn't this handled by the `@robot.brain.userForId()` method? idk, i'm confused.

src/bot.coffee
```diff
@@ -221,16 +229,14 @@ class SlackBot extends Adapter
       # If the reaction is to a message, then the item.channel property will contain a conversation ID
       # Otherwise reactions can be on files and file comments, which are "global" and aren't contained in a conversation
       user.room = event.item?.channel # when the item is not a message this will be undefined
+      # Convert item user into a Hubot user
+      item_user = @robot.brain.userForId event.item_user.id, event.item_user
```
aoberoi: what happens when there is no `item_user` property? wouldn't `event.item_user.id` cause an error at runtime (cannot find property `id` on undefined)? this is a use case when the reaction is on a message that was created by an Incoming Webhook.

src/client.coffee
```diff
@@ -169,28 +174,32 @@ class SlackClient
...
+          event.user = fetched.user
+          try @eventHandler(event)
+          catch error then @robot.logger.error "An error occurred while processing an RTM event: #{error.message}."
+        # User always preferred over bot
+        else if event.bot_id
+          # bot_id exists on all messages with subtype bot_message
+          # these messages only have a user property if sent from a bot user (xoxb token). therefore
+          # the above assignment will not happen for all messages from custom integrations or apps without a bot user
+          @loadUsers(event.bot_id, (err, res) =>
+            event.user = res
```
aoberoi: what happens if `res` is `undefined`? that's a possibility according to how `loadUsers` is implemented.

src/client.coffee
```diff
@@ -169,28 +174,32 @@ class SlackClient
...
+            try @eventHandler(event)
```
aoberoi: i'm not a fan of this code repetition. can you find a way to invoke `@eventHandler(event)` in just one place?

test/bot.coffee
```diff
@@ -46,33 +46,33 @@ describe 'Logger', ->
...
   it 'Should not send empty messages', ->
-    sentMessages = @slackbot.send {room: 'general'}, 'Hello', '', '', 'world!'
+    sentMessages = @slackbot.send {room: @stubs.channel.id}, 'Hello', '', '', 'world!'
     # Removes undefined (or unsent) messages from sentMessages
     _.remove(sentMessages, _.isUndefined)
     sentMessages.length.should.equal 2
 
   it 'Should not fail for inexistant user', ->
-    chai.expect(() => @slackbot.send {room: 'inexistant'}, 'Hello').to.not.throw()
+    chai.expect(() => @slackbot.send {room: 'U987'}, 'Hello').to.not.throw()
```
aoberoi: nitpicky again, but shouldn't this value look like a conversation ID (not a user ID)? or maybe that was what is intended, and then i think a comment could help clear that up. ditto for many of the edits below.

src/bot.coffee
shanedewael: That's a better way to state it, yeah. The `userForId()` method doesn't handle this because it requires a non-empty ID, which would not exist in this scenario.

test/bot.coffee
shanedewael: The `Client sending message` tests should be conversation ID and I changed it to reflect that, but some of the `Send messages` test are trying to send a DM to a user with no conversation ID. It was using `username` before, so now a few of those are using user ID. That being said, the current tests are limited because of lost functionality (`@client.openDM()`)

src/client.coffee
```diff
@@ -169,27 +196,34 @@ class SlackClient
...
+        # slackbot have the user property
+        if fetched.user
+          event.user = fetched.user
+          return event
+
+        else if event.bot_id
+          # bot_id exists on all messages with subtype bot_message
+          # these messages only have a user property if sent from a bot user (xoxb token). therefore
+          # the above assignment will not happen for all messages from custom integrations or apps without a bot user
+          Promise.promisify(@findBotUser, { context: @ })(event.bot_id).then((res) =>
```
aoberoi: I think you need `return` at the beginning of this line.

Issue #468, opened 2018-03-27
ReactionMessage item_user property should be marked as optional (Incoming Webhooks)


y-i:
### Description

hubot-slack can't react to `reaction_added` events if the message which reaction was added to is incoming webhook.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Reproducible in:

hubot-slack version: 4.4.0

node version: v8.9.4

OS version(s): macOS Sierra (10.12.6)

#### Steps to reproduce:

1. Join the bot which uses `robot.react()` to a channel.
2. Send a incoming webhook message to the channel.
3. Add reaction to the message.

#### Expected result:

`robot.react()` will be triggered.

#### Actual result:

`robot.react()` will not be triggered.

#### Attachments:

At [this point](https://github.com/slackapi/hubot-slack/blob/master/src/bot.coffee#L219), `item_user` is undefined in message if the message was posted by incoming webhook.
I can react to messages which were posted by incoming hook by changing `return unless user && item_user` to `return unless user`.

aoberoi:
thanks for the report! this _might_ be fixed in #465, because the code path changes significantly.

@y-i by any chance, is the user reacting to the incoming webhook message not a member of the local workspace? specifically, is this occurring in a shared channel?

@shanedewael do we have any tests that will cover this scenario? i don't think its specific to Incoming Webhook, but we should try.

y-i:
@aoberoi I didn't use a shared channel. So the user is a member of the local workspace.

aoberoi:
@y-i thanks, i was able to reproduce this (both on RTM and on Events API)

it turns out that this is expected (but not well documented) behavior. when the message is generated from an Incoming Webhook, there actually is no "user" responsible for sending it, so there's no `item_user` value at all.

[...]

PR #475, opened 2018-04-17
Fix Slack API Calls

shanedewael:
###  Summary

#465 introduced new calls to the Web API that were being incorrectly called and used, as noted in #468. This fixes the usage of the calls.

### DIFF
--- a/test/client.coffee
+++ b/test/client.coffee
@@ -27,8 +27,8 @@ describe 'onEvent()', ->
   it 'should emit pre-processed messages to the callback', (done) ->
     @client.onEvent (message) =>
       message.should.be.ok
-      #message.user.real_name.should.equal @stubs.user.real_name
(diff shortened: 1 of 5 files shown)


test/client.coffee
```diff
@@ -27,8 +27,8 @@ describe 'onEvent()', ->
   it 'should emit pre-processed messages to the callback', (done) ->
     @client.onEvent (message) =>
       message.should.be.ok
-      message.user.real_name.should.equal @stubs.user.real_name
-      message.channel.name.should.equal @stubs.channel.name
+      #message.user.real_name.should.equal @stubs.user.real_name
+      #message.channel.name.should.equal @stubs.channel.name
```
aoberoi: do we just not expect these to be equal anymore? can you explain why?

PR #483, opened 2018-04-25
WIP: optimize calls to conversations.info

aoberoi:
###  Summary

currently, `SlackClient` will fetch the conversation info for every incoming event, whether that object is used or not. this is wasteful because besides the conversation ID, the only time the info is used to inside `SlackTextMessage#buildText()` and that is to check the `is_im` property. this PR moves that fetch into `SlackTextMessage#buildText()` when its determined to be needed.

one idea to further optimize would be to cache the conversation info in memory (not in the Brain) for a specific and configurable TTL so that this fetch isn't executed as often.

no tests yet, this is to serve as an experiment for the team to discuss.


shanedewael:
Merged into #491

PR #485, opened 2018-05-03
Upgrade @slack/client to 4.2.0 to address CVE-2018-3728

bval:
###  Summary

CVE-2018-3728 applies to versions of hoek pre 4.2.1, which hubot-slack currently includes. This PR attempts to rectify that.

The 3.x branch of `@slack/client` has been abandoned and contains a locked reference to an older version of request which contains a vulnerable version of hoek. The only way forward is to move to the 4.x branch of `@slack/client` unless things change over there.

The migration guide can be found here:
https://github.com/slackapi/node-slack-sdk/wiki/Migration-Guide-for-v4

In order to get the suite to pass I had to s/RtmClient/RTMClient/g across this repo. I also had to modify 3 tests to mimic API changes upstream, but once done all tests pass and no source code modifications other than the capitalization of RTMClient were necessary.

This could use lots of eyes. Despite codecov showing 82% I am not familiar enough with this code to say what unexpected breakages may await.

### DIFF
--- a/docs/_pages/advanced_usage.md
+++ b/docs/_pages/advanced_usage.md
@@ -25,9 +25,9 @@ $ HUBOT_SLACK_TOKEN=xoxb-xxxxx HUBOT_SLACK_RTM_START_OPTS='{ "mpim_aware": true 
 
 ## Customizing the RTM Client
 
-The RTM connection is handled by the `RtmClient` class from our handy
+The RTM connection is handled by the `RTMClient` class from our handy
 **Slack Developer Kit for Node.js version 3**. By default, the adapter instantiates the client with the required
-`token` parameter, but more options are available. You can customize the options for the `RtmClient` instance by setting
+`token` parameter, but more options are available. You can customize the options for the `RTMClient` instance by setting
 an environment variable. The variable is called `HUBOT_SLACK_RTM_CLIENT_OPTS`, and its value should be a JSON-encoded
 string with the additional parameters as key-value pairs. Note that not every option can only be set to JSON-encodable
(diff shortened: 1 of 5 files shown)


aoberoi:
@bval thank you so much for bringing this to our attention.

unfortunately, we cannot simply update to the v4.x range of `@slack/client`, as that will break a number of APIs in this module. the largest breakage would be around the `RtmClient.dataStore` property, which simply does not exist in v4.x. Today's test suite is passing because we've been working on moving the implementation inside this module away from relying on `RtmClient.dataStore` for internal operations, but the fact exists that it is exposed to users of the package. That means this would have to be published as a semver:major change. once we get into that territory, i don't consider that a good fix, because most users would have to opt into this update specifically, and potentially migrate a huge amount of their own code to do so.

i've just chatted with the team and we have an alternative we'd like to move forward with. here's the summary: we publish a new *pre-release* version of the `@slack/client` package (which we also maintain) – lets call this `3.17.0-sec`. in this version, we update the `request` dependency to the latest. its important that it is distributed as a pre-release version because updating to this version would drop support for node versions < 4.x, and we don't want clients using `^3.x.x` as their version specifier to automatically pick this up and break (if they are running on these ancient versions). the next step is to update `hubot-slack` to depend strictly on `3.17.0-sec`. this should have the desired outcome of pulling in an updated `hoek` package into the dependency tree without breaking APIs.

wdyt?

aoberoi:
I've captured the steps in the referenced issues, so I'll close this PR.

bval:
@aoberoi 👍thanks for the summary and the action. Your solution makes a lot of sense, and of course you have the ability to bump `@slack/client` and I don't. 😄

PR #489, opened 2018-06-01
Decrease Web API Dependencies

shanedewael:
###  Summary

Add a local mapping for bot user ID to user representation to reduce dependency on `bots.info` and `users.info` calls.

Fetch event `user` and `item_user` from the brain if possible - if not, use `users.info`

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -37,6 +37,10 @@ class SlackClient
     # be removed without breaking this object's API. The property is no longer used internally.
     @format = new SlackFormatter(@rtm.dataStore, @robot)
 
+    # Map to convert bot user IDs (BXXXXXXXX) to user IDs (UXXXXXXXX/WXXXXXXXX) for events from custom 
+    # integrations and apps without a bot user
+    @botUserIdMap = {}
+
     # Event handling
     # NOTE: add channel join and leave events
     @rtm.on "message", @eventWrapper, this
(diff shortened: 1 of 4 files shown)


src/client.coffee
```diff
@@ -198,6 +202,28 @@ class SlackClient
...
+  ###*
+  # Fetch user info from the brain. If not available, call users.info
+  # @public
+  ###
+  fetchUser: (userId) ->
+    # User exists in the brain - retrieve this representation
+    return @robot.brain.data.users[userId] if @robot.brain.data.users[userId]?
+    
+    # User is not in brain - call users.info
+    # The user will be added to the brain in EventHandler
```
aoberoi: thanks for this comment! i was confused about whether it would be saved or not

src/client.coffee
```diff
@@ -229,13 +255,25 @@ class SlackClient
...
-              .then (res) =>
+          # fetched.bot will exist and be false if bot_id in @botUserIdMap 
+          # but is from custom integration or app without bot user
+          else if fetched.bot
+            # fetched.bot is user representation of bot since it exists in botToUserMap
+            if @botUserIdMap[event.bot_id]
+              event.user = fetched.bot
+
+            # bot_id exists on all messages with subtype bot_message
+            # these messages only have a user property if sent from a bot user (xoxb token). therefore
```
aoberoi: small typo: `these messages only have a user_id property`

PR #491, opened 2018-06-05
Update conversations.info calls to Web API

shanedewael:
###  Summary

Adds a basic cache for conversations and uses user representation in brain for mentions.

Now, conversations will only update if the stored information is older than 5 minutes. This means removing the logic of conversations in `client.coffee` and adding some basic logic into `message.coffee`. This will ensure mostly accurate conversation representations, but decrease the possibility of accidental rate limiting and unnecessary calls to conversations.info.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,10 @@
     {
       "email": "aoberoi@gmail.com",
       "name": "Ankur Oberoi"
+    },
+    {
+      "email": "sdewael@slack-corp.com",
+      "name": "Shane DeWael"
     }
   ]
 }
--- a/src/message.coffee
+++ b/src/message.coffee
@@ -96,8 +96,8 @@ class SlackTextMessage extends TextMessage
 
     # Replace links in text async to fetch user and channel info (if present)
     mentionFormatting = @replaceLinks(client, text)
-    # TODO: make this go through a conversations cache
-    fetchingConversationInfo = client.web.conversations.info(@_channel_id)
+    # Fetch conversation info
+    fetchingConversationInfo = @fetchConversation(client, @_channel_id)
     Promise.all([mentionFormatting, fetchingConversationInfo])
       .then (results) =>
         [ replacedText, conversationInfo ] = results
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -78,6 +78,10 @@
     {
       "email": "aoberoi@gmail.com",
       "name": "Ankur Oberoi"
+    },
+    {
+      "email": "sdewael@slack-corp.com",
+      "name": "Shane DeWael"
```
aoberoi: who do you think you are??

src/message.coffee
```diff
@@ -169,6 +185,23 @@ class SlackTextMessage extends TextMessage
       .then (substrings) ->
         return substrings.join("")
 
+  ###*
+  # Fetch conversation information from conversation map
+  # @private
+  ###
+  fetchConversation: (client, conversationId) ->
```
aoberoi: can this just be a method inside `client`? it feels a little strange to me that this message instance is responsible for reaching inside the `client` and updating its state (the `channelData` map).

src/message.coffee
```diff
@@ -95,23 +95,39 @@ class SlackTextMessage extends TextMessage
...
         text = replacedText
-        text = text.replace /&lt;/g, '<'
-        text = text.replace /&gt;/g, '>'
-        text = text.replace /&amp;/g, '&'
+        text = text.replace /&lt;/g, "<"
+        text = text.replace /&gt;/g, ">"
+        text = text.replace /&amp;/g, "&"
 
+        # Add conversation info to conversation map
+        if not client.channelData[@_channel_id]? and conversationInfo.channel?
```
aoberoi: i think adding the conversation info into the conversation map can be done in the same step as fetching the conversation info. in fact, as pointed out below, i think this should be responsibility of the client (the conversation map should be treated as private state and only the client knows that the data is cached).

src/message.coffee
```diff
@@ -179,6 +212,10 @@ class SlackTextMessage extends TextMessage
   # @returns {Promise<string>} - a string that can be placed into the text for this mention
   ###
   replaceUser: (client, id, mentions) ->
+    if client.robot.brain.data.users[id]?.name?
```
aoberoi: again, i think you should just call a method in the client, who can return a user, but the implementation of where that user comes from (either the brain or by calling `users.info`) is abstracted away from the message

aoberoi:
@shanedewael this is excellent work. hats off to you sir!

giphy

PR #506, opened 2018-07-10
improves checking for existence of strings before concatenation

### LINKED ISSUE #504: Message is not matching listener as expected
### Description

This issue is to track the problem first described in #500. In that issue, a bug was uncovered with the attachment formatting, and fixed. But @mininobu is still having a problem with a listener (created by the `hubot-broadlink-rm` script package) not matching messages it is meant to match. The key details are in [this comment](https://github.com/slackapi/hubot-slack/issues/500#issuecomment-402716037)

---

### Bug Report

#### Reproducible in:

hubot-slack version: 4.5.2

node version: 8.11.3

OS version(s): Raspbian 9.1（Raspberry Pi 3)

#### Steps to reproduce:

1. execute hubot bin/hubot --adapter slack
2. wait for hubot to connect to Slack RTM
3. Send a message from IFTTT
　　IFTTT "that" info
          Which channel? => Direct Message / @homebot
          Message             => send tv:on
     ※The message was sent from the Android application of IFTTT.
4. A message arrives from IFTTT to Slack (it is confirmed that it will be displayed by the Slack application)
5. The listener never executes (according to logs) and the RM Mini never receives the command.

#### Expected result:

RM Mini to operate.

#### Actual result:

RM Mini is idle.

#### Attachments:

```
DEBUG Received text message in channel: XXXXXXXXX, from: undefined (bot)
DEBUG No listeners executed; falling back to catch-all
```

aoberoi:
###  Summary

fixes #504

### DIFF
--- a/src/message.coffee
+++ b/src/message.coffee
@@ -87,12 +87,12 @@ class SlackTextMessage extends TextMessage
   ###
   buildText: (client, cb) ->
     # base text
-    text = @rawMessage.text
+    text = if @rawMessage.text? then @rawMessage.text else ""
 
     # flatten any attachments into text
     if @rawMessage.attachments
       attachment_text = @rawMessage.attachments.map((a) -> a.fallback).join("\n")
-      text = text + "\n" + attachment_text
+      text = text + "\n" + attachment_text if attachment_text?
 
     # Replace links in text async to fetch user and channel info (if present)
     mentionFormatting = @replaceLinks(client, text)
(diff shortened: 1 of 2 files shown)


src/message.coffee
```diff
@@ -87,12 +87,12 @@ class SlackTextMessage extends TextMessage
...
   buildText: (client, cb) ->
     # base text
-    text = @rawMessage.text
+    text = if @rawMessage.text? then @rawMessage.text else ""
 
     # flatten any attachments into text
     if @rawMessage.attachments
       attachment_text = @rawMessage.attachments.map((a) -> a.fallback).join("\n")
-      text = text + "\n" + attachment_text
+      text = text + "\n" + attachment_text if attachment_text.length > 0
```
shanedewael: Might be simpler to use `if !!attachment_text` here

aoberoi: i personally think this is more explicit and nicer. `attachment_text` is always going to be a string (thanks to `Array.prototype.join()`), its just that sometimes it will be an empty string. there's no other falsy value to be concerned with.

shanedewael: 👍

PR #520, opened 2018-08-01
One way to fix bot_not_found due to SlackBot

### LINKED ISSUE #519: Scripts triggered by SlackBot's /remind no longer working
### Description

In our slack, we had a trigger we wanted every day at a certain time, so used slack's `/remind #channel every day at 11am @hubot runScript` which worked great until Monday July 30th.  (We suspect Slack changed something on their end about how SlackBot is represented in its `user` object, and hubot-slack doesn't support some part of that)

Now it seems produce an error when that remind trigger runs:
`ERROR Incoming RTM message dropped due to error fetching info for a property: bot_not_found.`

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

---

### Bug Report

#### Reproducible in:

hubot-slack version: hubot-slack@4.5.3

node version: node 8.9.4 (npm 6.1.0)

OS version(s): CentOS Linux release 7.3.1611 (Core) 

#### Steps to reproduce:

1. Have a `robot.hear` command, say triggered on the text `command`, that does anything
2. Have hubot in a Slack channel, say called `#placeholder`
3. Enter the command `/remind #placeholder in 1 minute hubot command`
4. Wait 1 minute, in the channel you should see `slackbot [12:01 PM] Reminder: @hubot ping.`

#### Expected result:

The command result to occur

#### Actual result:

Nothing in the chat, error logged: `ERROR Incoming RTM message dropped due to error fetching info for a property: bot_not_found.`, from this line of code: https://github.com/slackapi/hubot-slack/blob/master/src/client.coffee#L364

#### Attachments:

N/A

Eibwen:
fixes #519
Specifically when /remind is setup to trigger a hubot event

###  Summary

Prevent `bot_not_found` when looking up SlackBot using `bots.info` with id:B01 from causing an error that stops execution of listeners

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -225,6 +225,7 @@ class SlackClient
   # @public
   ###
   fetchBotUser: (botId) ->
+    return Promise.resolve(false) if (botId == "B01")
     return Promise.resolve(@botUserIdMap[botId]) if @botUserIdMap[botId]?
 
     # Bot user is not in mapping - call bots.info
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -39,7 +39,9 @@ class SlackClient
 
     # Map to convert bot user IDs (BXXXXXXXX) to user representations for events from custom
     # integrations and apps without a bot user
-    @botUserIdMap = {}
+    @botUserIdMap = {
+      "B01": { id: "B01", user_id: "USLACKBOT" }
+    }
 
     # Map to convert conversation IDs to conversation representations
     @channelData = {}
@@ -225,7 +227,6 @@ class SlackClient
   # @public
   ###
   fetchBotUser: (botId) ->
-    return Promise.resolve(false) if (botId == "B01")
     return Promise.resolve(@botUserIdMap[botId]) if @botUserIdMap[botId]?
 
     # Bot user is not in mapping - call bots.info
--- a/test/client.coffee
+++ b/test/client.coffee
@@ -312,10 +312,11 @@ describe 'fetchBotUser()', ->
     result = @client.fetchBotUser @stubs.bot.id
     result.should.be.Promise()
 
-  it 'should return false if id is slackbots id', ->
+  it 'should return constant data if id is slackbots id', ->
     @client.fetchBotUser @stubs.slack_bot.id
     .then((res) ->
-      res.should.equal false
+      res.id.should.equal @stubs.slack_bot.id
+      res.user_id.should.equal @stubs.slack_bot.user_id
     )
 
 describe 'fetchUser()', ->
--- a/test/client.coffee
+++ b/test/client.coffee
@@ -312,6 +312,12 @@ describe 'fetchBotUser()', ->
     result = @client.fetchBotUser @stubs.bot.id
     result.should.be.Promise()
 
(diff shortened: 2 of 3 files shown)


aoberoi:
Thanks for taking the time to offer a solution!

I'll have to confirm this, but I think we know that slackbot always has bot ID `B01` **and** user ID `USLACKBOT`. If that's true, the solution I'd rather merge would be changing [this line](https://github.com/slackapi/hubot-slack/blob/c84f7f748afd4e4a3b0ef1c26fc234a68c8de427/src/client.coffee#L318) to this:

```coffeescript
      if event.bot_id?
        fetches.bot = if event.bot_id == "B01" then { user_id: "USLACKBOT" } else @fetchBotUser event.bot_id
```

Then, when the code reaches [this line](https://github.com/slackapi/hubot-slack/blob/c84f7f748afd4e4a3b0ef1c26fc234a68c8de427/src/client.coffee#L342), the fetch for the user representation would be made and resolved to a real user object (instead of just false).

Eibwen:
Ah that is likely a better solution.

I didn't look too closely into how `@botUserIdMap` worked, but saw that some of the values were just `false`, so went with that.  Another possible option would be to insert a `B01` entry into that lookup upon creation or something

It seems like `slackbot` from the `users.list` endpoint on a small slack group I'm in at least is currently including `"is_bot": false,` which is quite surprising to me.  So that was also part of why I just went with `false` (but really didn't put much thought into it :) )

Eibwen:
Also happen to know if the CLAassistant is based on username, or commit email, or both?  (I was extra lazy and edited each file in the browser, so I'm not even sure what email these commits were created with, but wondering if that is the mis-match)  Assuming that is the mismatch, shall I just sign the CLA again with whatever probably the eibwen@github email that these were likely created with?

Or does it just take a while to update?

aoberoi:
> Another possible option would be to insert a B01 entry into that lookup upon creation or something

I like that even better! We can just hardcode one entry into the map when its initialized [here](https://github.com/slackapi/hubot-slack/blob/c84f7f748afd4e4a3b0ef1c26fc234a68c8de427/src/client.coffee#L42).

> It seems like slackbot from the users.list endpoint on a small slack group I'm in at least is currently including "is_bot": false, which is quite surprising to me.

Slackbot is... a special snowflake ❄️. Ideally, yes, it makes more sense for slackbot to have `is_bot: "true"`, but practically I don't think there should be much of a difference. I don't think its worth patching in this Hubot SDK and then supporting that patch for the conceivable future, unless we can think of a reason this would make a script behave in unintended ways.

Eibwen:
Ah, I didn't click the badge the first time.  Whatever I clicked just took me to the google form without requiring the GitHub App authorization, maybe from https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md ?

Idk worked great when I clicked the badge!  Did also change the email to the email the commits have, but with the App authorization I imagine it has access to all the emails on my github account

aoberoi:
Whoohoo! This is great. Thank you so much.

PR #525, opened 2018-08-25
Fix tests on new hubot

mistydemeo:
###  Summary

This fixes running the tests with Hubot 3.x.

I was seeing these four `instanceof` tests fail. I believe the reason for this is that Hubot 3.x is written in JavaScript using ES2015 classes, and there's a subtle issue with `instanceof` not playing nicely when CoffeeScript classes enter the picture.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,9 @@ node_js:
   - "5"
   - "4"
   - "4.2"
+env:
+  - HUBOT_VERSION=">=2.0 <3"
+  - HUBOT_VERSION="^3.0"
 after_success:
   - npm run codecov
+script: npm install && npm install hubot@"$HUBOT_VERSION" && npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,8 @@ env:
   - HUBOT_VERSION="^3.0"
 after_success:
   - npm run codecov
+# The second step here installs the specific hubot version determined
+# by the `env` specified in the matrix. That lets us test against
+# multiple Hubot versions. This is a bit messy, but works around
+# package.json only being able to specify one version.
 script: npm install && npm install hubot@"$HUBOT_VERSION" && npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,5 @@ node_js:
   - "5"
   - "4"
   - "4.2"
-  - "0.12"
 after_success:
   - npm run codecov
--- a/package.json
+++ b/package.json
@@ -29,14 +29,14 @@
     "lodash": "^4.17.10"
   },
   "peerDependencies": {
-    "hubot": "^2.0.0"
+    "hubot": ">= 2.0.0"
   },
   "devDependencies": {
     "chai": "^3.5.0",
     "coffee-coverage": "^2.0.0",
     "coffeescript": "^1.12.7",
     "codecov": "^2.3.1",
-    "hubot": "^2.19.0",
+    "hubot": ">= 2.19.0",
     "istanbul": "^0.4.3",
     "mocha": "^3.5.3",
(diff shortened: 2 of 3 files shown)


aoberoi:
@mistydemeo my only reservation with this PR is that we will no longer have tests that run against `hubot@2` (since npm will always resolve to the latest version of `hubot` with the change in range specifier). ideally, we have more entries in the travis test matrix for all combinations of supported hubot major versions and node versions.

regarding the failure: i don't like this, but i'm willing to silently drop support for node v0.12. there doesn't seem to be a reasonable alternative. here are some that i've considered:
1. release this in a major version that formally drops support for node versions < 6. the problem is that we have no published support timeline that would have called for this and we have _many_ more changes we'd like to do in a major version.
2. never ever release a new version. clearly, not a good solution, it wouldn't even guard users who choose to upgrade hubot on their own.
3. silently drop support for node v0.12 without bumping the major version number of this package. if someone was running on this ancient version of node, they would suddenly see their service go down. i don't have any data to query to see how popular this is. however, the silver lining is that programs running this ancient version of node are long overdue to upgrade their node version because they are using something with known security issues that have not been patched. this would be a forcing function on changing, and likely just involves a simple download/install to resolve.

aoberoi:
Just noting down that @mistydemeo has agreed to add another dimension to the test matrix to cover the variations in `hubot` versions. 🙌

aoberoi:
related #529

titaniumbones:
Hi, I'm just wondering about the status of this PR -- is it likely to be merged soon or are there concerns about it htat I can't see in the issue? Hoping to run a Hubot 3.x bot but hesitant to deploy before the official hubot-slack adapter supports it!

aoberoi:
I’m happy to merge this change as soon as we have test coverage across both 2.x and 3.x versions of hubot. @mistydemeo, any comment around when we should expect that?

mistydemeo:
Sorry for the delay; I think I've got the Travis matrix taken care of, by using `env` to set a Hubot version and then manually `npm install` that version to override the version that would otherwise get selected.

aoberoi:
@mistydemeo ah, that's pretty clever, but 😭 i don't think it worked the way we intend it to.

check out the build log for the job testing hubot 2.x with node 10: https://travis-ci.org/slackapi/hubot-slack/builds/434342709

screen shot 2018-09-29 at 8 46 17 am

Looks like it still got hubot 3.x

mistydemeo:
We end up installing Hubot twice! Which is awkward, but it's an artifact of the fact the `package.json` only resolves to one version, and we have different versions to test with. The main `npm install` will pick 3.x, but then we replace it with `npm install hubot@"$HUBOT_VERSION"` in the script before running `npm test`. So while `npm list | grep hubot` after the main install will always be the same version, `npm install hubot@">=2.0 <3" && npm list | grep hubot` should always show 2.x and `npm install hubot@"^3.0" && npm list | grep hubot` should always show 3.x.

.travis.yml
```diff
@@ -8,5 +8,9 @@ node_js:
   - "5"
   - "4"
   - "4.2"
+env:
+  - HUBOT_VERSION=">=2.0 <3"
+  - HUBOT_VERSION="^3.0"
 after_success:
   - npm run codecov
+script: npm install && npm install hubot@"$HUBOT_VERSION" && npm test
```
aoberoi: doh, that makes total sense. thanks for walking me through it. maybe we can add a comment here so that future maintainers have some indication of how this works. something along the lines of "explicitly install hubot with an overridden version so that we can check for issues before the latest major version as seen in package.json". that might be too wordy, but something like that would be good.

mistydemeo: OK, pushed with a comment!

Issue #526, opened 2018-09-06
Coffeescript 2.x compatibility


mistydemeo:
## Description

Describe your issue here.

## What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

## Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

## Introduction

The primary incompatibility between Coffeescript 1.x and 2.x is the class type. Coffeescript 1.x's object model is a custom one which compiles down to JS prototype-based objects, since Coffeescript 1.x predates JS having its own class model. Coffeescript 2.x classes instead compile down to ECMAScript 2015 classes in order to improve interoperability with modern JS. This required two major backwards-incompatible changes in order to comply with JS class requirements:

1. `super` is now both a keyword *and* a method. This means that `super`, when called as a method, has to be called as `super()` in order to disambiguate.
2. ECMAScript 2015 requires that `super` always be called before attributes are read or assigned within a constructor, so Coffeescript now has the same requirement. This means that constructors using the `@ivar` attribute name/assignment shorthand must call `super()` immediately before doing anything else.

While Hubot 3.x is compatible with Coffeescript 2.x, since Hubot itself is no longer written in CS, hubot-slack isn't yet compatible. I've outlined the compatibility testing I've performed with the problems I've identified.

### Hubot 2.x + Coffeescript 1.x

:white_check_mark: Officially supported and compatible.

### Hubot 3.x + Coffeescript 1.x

:white_check_mark: Officially supported and compatible.

### Hubot 2.x + Coffeescript 2.x

:no_entry_sign: Not supported and not compatible. Hubot 2.x classes are written in Coffeescript before the release of 2.x.

### Hubot 3.x + Coffeescript 2.x

:no_entry_sign: Supported, but not compatible with hubot-slack.

It looks like there's probably only one compatibility issue, but I need to test more:

[...]

PR #528, opened 2018-09-06
WIP: add compatibility for Coffeescript 2

mistydemeo:
###  Summary

This PR starts the work on making sure the classes in this code work in Coffeescript 2. As mentioned previously, a few of the semantic differences between Coffeescript 1 classes and ES2015 classes has introduced some backwards compatibility issues, documented here: https://coffeescript.org/#breaking-changes

Currently, this mostly works, but trying to actually launch Hubot fails with a `TypeError: Cannot read property 'bind' of undefined` error. I believe this *may* be because of the following:

> ES2015 classes don’t allow bound (fat arrow) methods. The CoffeeScript compiler goes through some contortions to preserve support for them, but one thing that can’t be accommodated is calling a bound method before it is bound

refs #526.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "devDependencies": {
     "chai": "^3.5.0",
     "coffee-coverage": "^2.0.0",
-    "coffeescript": "^1.12.7",
+    "coffeescript": "^2.3.1",
     "codecov": "^2.3.1",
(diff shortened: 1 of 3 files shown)


seratch:
Let me close this as #565 has been merged into master branch

PR #440, opened 2017-09-26
Add callback support to bot::{reply,send}

alFReD-NSH:
* [x] I've read and understood the [Contributing guidelines](./CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](./CODE_OF_CONDUCT.md).
* [x] I've been mindful about doing atomic commits, adding documentation to my changes, not refactoring too much.
* [x] I've a descriptive title and added any useful information for the reviewer. Where appropriate, I've attached a screenshot and/or screencast (gif preferrably).
* [x] I've written tests to cover the new code and functionality included in this PR.
* [x] I've read, agree to, and signed the [Contributor License Agreement (CLA)](https://docs.google.com/a/slack-corp.com/forms/d/1q_w8rlJG_x_xJOoSUMNl7R35rkpA7N6pUkKhfHHMD9c/viewform).

#### PR Summary
This let's `SlackBot::reply` and `SlackBot::send` take an optional callback. While this is not documented, if a callback is passed to Hubot's `Response::send` and `Response::reply`, [Hubot will pass it around.](https://github.com/hubotio/hubot/blob/v3.0.1/src/response.js#L113-L115).

My use case is that, I want to get the `ts` of the message that is sent and then update it later. 

#### Related Issues
N/A

#### Test strategy
Passes the mocha callback to send and reply, so the test will only pass if the callback is called, also checking output to make sure the callback is not related, so not breaking compatibility in any way.

PR #540, opened 2018-10-15
Add callback support to SlackBot::{send,reply}

rtlechow:
###  Summary

Allows `SlackBot::{send,reply}` to receive an optional callback. While
this is not documented, if a callback is passed to Hubot's
`Response::{send,reply}`, Hubot will pass it around.

Shamelessly plagiarized from #440 because I'm in need of this
functionality. Fixed the tests to work with `@stubs`.


rtlechow:
BTW, if you sign the CLA prior to creating a PR (by following the link from the PR template), the status doesn't get checked. 🤓

package.json
```diff
@@ -26,7 +26,8 @@
   "dependencies": {
     "@slack/client": "3.16.1-sec.2",
     "bluebird": "^3.5.1",
-    "lodash": "^4.17.10"
+    "lodash": "^4.17.10",
+    "es6-promise": "~4.2.5"
```
aoberoi: i don't see why this dependency is required. bluebird is already a dependency, does it not meet  the needs?

src/bot.coffee
```diff
@@ -2,6 +2,7 @@
 {SlackTextMessage, ReactionMessage, PresenceMessage}                              = require "./message"
 SlackClient                                                                       = require "./client"
 pkg                                                                               = require "../package"
+Promise = global.Promise || require('es6-promise')
```
aoberoi:
```suggestion
Promise = require("bluebird");
```

src/bot.coffee
```diff
@@ -80,10 +81,13 @@ class SlackBot extends Adapter
...
   ###
   send: (envelope, messages...) ->
-    # TODO: if the sender is interested in the completion, the last item in `messages` will be a function
+    callback = ->
+    if typeof(messages[messages.length - 1]) == 'function'
+      callback = messages.pop()
     for message in messages
       # NOTE: perhaps do envelope manipulation here instead of in the client (separation of concerns)
       @client.send(envelope, message) unless message is ""
+    Promise.all(messages).then(callback.bind(null, null), callback)
```
aoberoi: i'm not sure how this works. `messages` is an array of either a string or an object, which might have a callback as the last item (after this change). i think what you meant is to collect all the Promises returned from `@client.send(...)`. maybe the for loop needs to be refactored into a something like `messagePromises = messages.map(...)`, so that this line can become `Promise.all(messagePromises)`. you also need to be careful in the `messages.map(...)` mapping function to skip operating on any item that looks like a function, probably by just returning a simple value or `Promise.resolve()`.

src/bot.coffee
```diff
@@ -92,12 +96,15 @@ class SlackBot extends Adapter
...
-    # TODO: if the sender is interested in the completion, the last item in `messages` will be a function
+    callback = ->
+    if typeof(messages[messages.length - 1]) == 'function'
+      callback = messages.pop()
     for message in messages
       if message isnt ""
         # TODO: channel prefix matching should be removed
         message = "<@#{envelope.user.id}>: #{message}" unless envelope.room[0] is "D"
         @client.send envelope, message
+    Promise.all(messages).then(callback.bind(null, null), callback)
```
aoberoi: ditto with above.

aoberoi:
thanks for the contribution. i'm happy to merge this and gain a new contributor! there's just a couple comments that need to be addressed first.

rtlechow:
@aoberoi Finally got around to fixing it up. Any thoughts on improving coverage?

src/bot.coffee
```diff
@@ -84,10 +84,11 @@ class SlackBot extends Adapter
     callback = ->
     if typeof(messages[messages.length - 1]) == 'function'
       callback = messages.pop()
-    for message in messages
+    messagePromises = messages.map (message) =>
+      return Promise.resolve() if typeof(message) is 'function'
```
aoberoi: small style nit, we're trying to use double quotes everywhere.

src/bot.coffee
```diff
@@ -99,12 +100,13 @@ class SlackBot extends Adapter
...
```
aoberoi: same as above.

seratch:
As @aoberoi already approved this PR before, it looks good to me, too. We took a bit long time after this PR was created. Just in case, I'll wait for responses from others until next week before merging this.

Issue #537, opened 2018-10-10
robot.react robot.hearReaction not a function error with hubot 3.1.1


iancward:
### Description

Describe your issue here.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Reproducible in:

hubot-slack version: 4.5.5

node version: 8.12.0

OS version(s): Fedora 27

#### Steps to reproduce:

1. Use hubot-slack version 4.5.5 and hubot version 2.19.0, with a script that has `robot.react` or `robot.hearReaction` call
2. Confirm bot working with these versions
3. Upgrade hubot to version 3.1.1
4. Start bot, receive `TypeError: robot.react is not a function` or `TypeError: robot.hearReaction is not a function` error on bot startup

#### Expected result:

Hubot should not fail with this error on upgrade, or proper usage of `robot.react` / `robot.hearReaction` for Hubot 3.x should be properly documented here

#### Actual result:

Hubot threw that error, dumped a stack trace, and aborted at startup.

aoberoi:
oh no, sorry about that. it seems we haven't properly considered and tested this functionality on both Hubot 3.x and 2.x.

my guess is that the way in which we extend the prototype for classes in Coffeescript doesn't jive with ES2015 classes. In Coffeescript, we depend on the `Class::methodName = fn` syntax to add methods to the prototype of `Class`. see example:

https://github.com/slackapi/hubot-slack/blob/a4421541b308d475aa43274d41e17bd2020cbcd0/src/extensions.coffee#L13

but briefly looking at the documentation, it seems like this syntax should just add to the prototype the way we expect. we'll have to spend some more time reproducing this and debugging what's actually going on here.

aoberoi:
[...]

PR #541, opened 2018-10-22
Imports Hubot prototype dynamically for extensions

### LINKED ISSUE #538: robot.presenceChange never triggered
### Description

I'm trying to use robot.presenceChange to detect when user comes online and send him a message. I've tried changing my status to away and logging out of slack, but the event never triggers.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ x] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

---

### Bug Report

#### Reproducible in:

hubot-slack version: 4.4.5
node version: v8.10.0

#### Steps to reproduce:

1. Start hubot with the attached script
2. Change your slack status
3. Console window will be clear as a night sky

#### Expected result:

I expect my code to be executed every time I change my slack status to away or active.

#### Actual result:

Code never executes.

#### Attachments:
This is the code from basic_usage.md with console.log res added at the begginig. 

`module.exports = (robot) ->
    robot.presenceChange (res) ->
        console.log res
        # res.message is a PresenceMessage instance that represents the presence change Hubot just heard
        names = (user.name for user in res.message.users).join ", "
        message = if res.message.presence is "away" then "Bye bye #{names}" else "Glad you are back #{names}"
        robot.logger.debug message`

shanedewael:
###  Summary

Hubot `v3.0.0` updated the source to `es2015`, which includes implementing a coffeescript-compatible wrapper to not break adapters. However, this meant that we were adding our extensions (`hearReaction` and `presenceChange`) to the coffeescript wrapper rather than the `Robot` prototype. This fixes that by checking `robot.name` (which is `CoffeeScriptCompatibleClass` for Hubot `v3.0.0` or higher) and importing the proper hubot based on that.

Fixes #538 and #537 

### DIFF
--- a/test/bot.coffee
+++ b/test/bot.coffee
@@ -1,27 +1,32 @@
 should = require 'should'
 chai = require 'chai'
-{ EnterMessage, LeaveMessage, TopicMessage, CatchAllMessage, Robot } = require.main.require 'hubot'
+mockery = require 'mockery'
+hubotSlackMock = require '../slack.coffee'
+mockery.registerMock('hubot-slack', hubotSlackMock);
+
+{ EnterMessage, LeaveMessage, TopicMessage, CatchAllMessage, Robot, loadBot } = require.main.require 'hubot'
 { SlackTextMessage, ReactionMessage, PresenceMessage } = require '../src/message'
 SlackClient = require '../src/client'
 _ = require 'lodash'
 
 describe 'Adapter', ->
+  before ->
+    mockery.enable({
+      warnOnUnregistered: false
+    });
+  
+  after ->
+    mockery.disable()
+  
   it 'Should initialize with a robot', ->
     @slackbot.robot.should.eql @stubs.robot
 
-  it 'Should add the `react` method to the hubot `Robot` prototype', ->
-    Robot.prototype.react.should.be.an.instanceOf(Function).with.lengthOf(3)
-
-    # This is a sanity check to ensure the @slackbot.robot stub is proper.
-    @slackbot.robot.listen.should.be.an.instanceOf(Function).with.lengthOf(3)
-    @slackbot.robot.react.should.be.an.instanceOf(Function).with.lengthOf(3)
-
-  it 'Should add the `presenceChange` method to the hubot `Robot` prototype', ->
-    Robot.prototype.presenceChange.should.be.an.instanceOf(Function).with.lengthOf(3)
-
-    # This is a sanity check to ensure the @slackbot.robot stub is proper.
-    @slackbot.robot.listen.should.be.an.instanceOf(Function).with.lengthOf(3)
-    @slackbot.robot.presenceChange.should.be.an.instanceOf(Function).with.lengthOf(3)
+  it 'Should load an instance of Robot with extended methods', ->
+    loadedRobot = loadBot('', 'slack', false, 'Hubot')
+    
+    # Check to make sure presenceChange and react are loaded to Robot
+    loadedRobot.presenceChange.should.be.an.instanceOf(Function).with.lengthOf(3)
+    loadedRobot.react.should.be.an.instanceOf(Function).with.lengthOf(3)
 
 describe 'Connect', ->
   it 'Should connect successfully', ->
--- a/package.json
+++ b/package.json
@@ -33,12 +33,13 @@
   },
   "devDependencies": {
     "chai": "^3.5.0",
+    "codecov": "^2.3.1",
     "coffee-coverage": "^2.0.0",
     "coffeescript": "^1.12.7",
-    "codecov": "^2.3.1",
     "hubot": ">= 2.19.0",
     "istanbul": "^0.4.3",
     "mocha": "^3.5.3",
+    "mockery": "^2.1.0",
     "should": "^8.0.0"
   },
   "engines": {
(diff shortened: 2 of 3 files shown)


aoberoi:
@shanedewael the implementation looks good, but i'd feel a lot better merging if we had a test that failed on the previous revision and passes using this code. is that something you've already looked into?

aoberoi:
okay i've been doing a bit more research into this problem, so that we can get a better understanding of this solution and hopefully how we might test this solution. here's what i've got.

in the Hubot v3.x source, the following code appears:

```javascript
function makeClassCoffeeScriptCompatible (klass) {
  function CoffeeScriptCompatibleClass () {
    const Hack = Function.prototype.bind.apply(klass, [null].concat([].slice.call(arguments)))
    const instance = new Hack()

    // pass methods from child to returned instance
    for (const key in this) {
      instance[key] = this[key]
    }

    // support for constructor methods which call super()
    // in which this.* properties are set
    for (const key in instance) {
      this[key] = instance[key]
    }

    return instance
  }
  inherits(CoffeeScriptCompatibleClass, klass)

  return CoffeeScriptCompatibleClass
}
```

ES2015 classes constructor functions such as `Robot` are each passed to `makeClassCoffeeScriptCompatible()` as the `klass` argument. The return value of this function is then exported as the class exported from the package, in this case as `Robot`. Let's dissect this to see why this isn't working before the change in this PR.

The return value, `CoffeeScriptCompatibleClass`, is a function that is meant to be used as a constructor function to stand in place of `klass`. That function is wired up to "inherit" from `klass`. What does this type of "inherit" mean? Let's take a look at the code for that.

```javascript
/**
 * Inherit the prototype methods from one constructor into another.
 *
 * The Function.prototype.inherits from lang.js rewritten as a standalone
 * function (not on Function.prototype). NOTE: If this file is to be loaded
 * during bootstrapping this function needs to be rewritten using some native
 * functions as prototype setup using normal JavaScript does not work as
 * expected during bootstrapping (see mirror.js in r114903).
 *
 * @param {function} ctor Constructor function which needs to inherit the
 *     prototype.
 * @param {function} superCtor Constructor function to inherit prototype from.
 * @throws {TypeError} Will error if either constructor is null, or if
 *     the super constructor lacks a prototype.
 */
function inherits(ctor, superCtor) {


  if (ctor === undefined || ctor === null)
    throw new ERR_INVALID_ARG_TYPE('ctor', 'Function', ctor);


  if (superCtor === undefined || superCtor === null)
    throw new ERR_INVALID_ARG_TYPE('superCtor', 'Function', superCtor);


  if (superCtor.prototype === undefined) {
    throw new ERR_INVALID_ARG_TYPE('superCtor.prototype',
                                   'Function', superCtor.prototype);
  }
  Object.defineProperty(ctor, 'super_', {
    value: superCtor,
    writable: true,
    configurable: true
  });
  Object.setPrototypeOf(ctor.prototype, superCtor.prototype);
}
```

to be continued...

aoberoi:
After much reasoning and research, I was above to build the following model of the objects in memory when using `hubot@3` in combination with this package (before the change in this PR).

coffeescriptcompatibleclasses

All the various relationships are drawn, but I'll cut to the punchline:

It is because the `Robot` instance that all scripts have access to, called `robotSingleton` in the model above, is [instantiated through the `loadBot()` helper](https://github.com/hubotio/hubot/blob/85284750c878205aeae7338513961b29f0a1d18f/es2015.js#L27) that the instance has no access to the extension methods. If the `Robot` instance had been created using the `CoffeeScriptCompatibleRobot` class (as its called in the model above), or otherwise the `Robot` export as seen from the point of view of this adapter, then there would be access to those methods - that would look like `instance` in the model.

However, adapters don't have the opportunity to change how the Hubot framework initialized the Robot instance.

The change in this PR effects this state by placing the extension methods on the rightmost `Anonymous` object, which in turn is accessible to `robotSingleton`. That is how this fix works.

Now, knowing this, we can come up with an adequate test. I think the test should call `loadBot()` (from the package exports) which should return an instance of `Robot`, and then call one of the extension methods on that object. As long as the test doesn't throw, we should know that our implementation works.

test/bot.coffee
```diff
@@ -1,27 +1,32 @@
...
+
+{ EnterMessage, LeaveMessage, TopicMessage, CatchAllMessage, Robot, loadBot } = require.main.require 'hubot'
 { SlackTextMessage, ReactionMessage, PresenceMessage } = require '../src/message'
 SlackClient = require '../src/client'
 _ = require 'lodash'
 
 describe 'Adapter', ->
+  before ->
+    mockery.enable({
+      warnOnUnregistered: false
```
aoberoi: out of curiosity, why is this set to `false`?

aoberoi:
LGTM! I just had a question out of curiosity, but otherwise I feel good about landing this :)

test/bot.coffee
shanedewael: Unregistered means all of the `requires` we aren't mocking, so every package used in the tests  prints a warning. This makes it so that won't happen.

frandallfarmer:
I'm confused by this checkin as a fix for #538 

What is the correct syntax for handling presenceChange events now?

The sample code on [pages](https://github.com/slackapi/hubot-slack/blob/master/docs/_pages/basic_usage.md)
still does not function (still refers to robot.presenceChange).

What is the correct syntax now, and should I file a bug to fix the docs to match?

PR #556, opened 2019-03-25
update dev dependencies

Trott:
This updates the dev dependencies just enough to get a clean `npm
audit`.

###  Summary

mocha and codecov dev dependencies are out of date, resulting in scary warnings from recent versions of npm. This fixes those issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,12 +33,12 @@
   },
   "devDependencies": {
     "chai": "^3.5.0",
-    "codecov": "^2.3.1",
+    "codecov": "^3.2.0",
     "coffee-coverage": "^2.0.0",
     "coffeescript": "^1.12.7",
     "hubot": ">= 2.19.0",
     "istanbul": "^0.4.3",
-    "mocha": "^3.5.3",
+    "mocha": "^4.1.0",
     "mockery": "^2.1.0",
     "should": "^8.0.0"
   },


aoberoi:
thanks @Trott!

just curious, are you making any changes in particular? this repo is waning on activity recently, so we should prob chat if you plan on contributing anything significant. there's also other projects that might already be better solutions and might already work the way you need.

Trott:
> just curious, are you making any changes in particular? this repo is waning on activity recently, so we should prob chat if you plan on contributing anything significant. there's also other projects that might already be better solutions and might already work the way you need.

I was going to update it so it doesn't use the deprecated Slack data store thing, but if there's a different project that I should be using, I'm happy to migrate.

aoberoi:
>  was going to update it so it doesn't use the deprecated Slack data store thing

we took a look at this a little while back and already removed any internal use of the data store, from within the adapter. when it comes to updating the updating `@slack/client` to a version that's actually removed the data store, we punted.

https://github.com/slackapi/hubot-slack/blob/e3acc28f224cb8fb60b2288b1395c25c063b27b2/src/client.coffee#L26-L28

The adapter (`SlackBot` instance) has a reference to the data store and is accessible to users (`robot.adapter.client.web.dataStore`). scripts might be depending on some of that data. that means upgrading the `@slack/client` dependency to something beyond the `v3.x` would be a breaking change (the `dataStore` is dropped in `v4.0.0` and beyond). we're not too keen on releasing a breaking change at the moment (we should probably decide how long we plan to support the current and future versions before we ship that).

if you have other thoughts, needs, ideas, rants, etc... let us know.

PS. thanks for all the work on node. its... not bad. 😉

PR #562, opened 2019-04-30
fix issue with using a string from env vars as a number

aoberoi:
###  Summary

In #560, we failed to convert the value from a string to a number, so that it could be used as the TTL. This PR fixes that problem.


aoberoi:
@mistydemeo good catch! going to update the PR to account for `NaN`.

PR #566, opened 2019-05-29
Adding a code of conduct file

colmdoyle:


seratch:
Sorry for the teams' belated response here. As we have the same one for other projects under this org, we should merge this. Thanks 👍

Issue #536, opened 2018-10-04
Request: allow detecting when hubot-slack has disconnected and will not reconnect


mistydemeo:
### Description

While hubot-slack has support for reconnecting when the connection to Slack is lost, there is a limit to the number of reconnection attempts. If several reconnection attempts fail, the bot will simply hang without a connection. This causes trouble for our process manager, since the client staying open means we don't end up automatically restart it. When this has happened for us a few times lately, we had to manually notice and restart the service in order to get it back up.

I'd like to see some form of event be emitted in this condition to allow us to determine when this has happened so that we can terminate the process and let the process manager restart it.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Reproducible in:

hubot-slack version: 4.5.5

node version: v8.2.0

OS version(s): Debian jessie

aoberoi:
related: https://github.com/slackapi/hubot-slack/pull/215

it looks like at some point in this package's history, there was an environment variable used to configure whether the process would terminate in this condition, not just fire an event. should we reintroduce that configuration? fire the event? both?

ben-nat-wallis:
It would be really nice to get this back into Hubot. We have a major issue at the moment with our hubot just disconnecting randomly from our slack, and there is no way to detect this so that we can restart the docker container it's running on.

charliekump-wf:
@ben-nat-wallis 

For our hubot, I workaround at the moment with a cron script that checks the logs for the reconnect message. Then kill/restart the container 🤷‍♂ 

```CHECK_STATUS=`docker logs $CONTAINER_ID | grep "INFO Slack client closed, waiting for reconnect" | wc -l````

mistydemeo:
There's been no activity because it was never fixed. I opened this in 2018.

seratch:
[...]

PR #572, opened 2019-07-29
Bot: catch and react to disconnection event

mistydemeo:
###  Summary

As described in #536, hubot-slack sometimes appears to disconnect and never come back. There is a [`disconnect` event](https://github.com/slackapi/node-slack-sdk/blob/13248f403758d7158c1bafe5db009e3f0416b4b8/lib/clients/events/client.js#L26-L27) which occurs specifically when the RTM client has disconnected and will not try again; this is currently not handled by the codebase, so that would explain why what we're seeing is going uncaught. In this PR, I've added a new handler for that method; right now I'm having that close the bot entirely, allowing the running process manager to clean up and restart Hubot.

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -37,6 +37,7 @@ class SlackBot extends Adapter
     # SlackClient event handlers
     @client.rtm.on "open", @open
     @client.rtm.on "close", @close
+    @client.rtm.on "disconnect", @disconnect
     @client.rtm.on "error", @error
     @client.rtm.on "authenticated", @authenticated
     @client.onEvent @eventHandler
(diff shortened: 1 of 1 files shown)


aoberoi:
@mistydemeo works for an organization that has signed the corporate CLA, so the individual CLA check can be safely ignored.

aoberoi:
CodeCov has aggregated the coverage report for this PR here: https://codecov.io/gh/slackapi/hubot-slack/pull/572

The GitHub / Travis CI integration does not seem to be working as expected. The coverage is satisfactory, and we're clear to merge.

PR #574, opened 2019-08-21
Use Coffeescript conditional assignment shorthand

sinisterstuf:
This is a really minor syntax-only change and shouldn't affect behaviour at all.

Because what we're assigning in the happy side of this condition is the same as what we're checking in its expression we can use Coffeescript's shorthand with the existential operator, i.e. instead of

    a = if b? then b else c

use this:

    a = b ? c

Source: https://coffeescript.org/#existential-operator

###  Summary

Describe the goal of this PR. Mention any related Issue numbers.

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -238,7 +238,7 @@ class SlackBot extends Adapter
 
       # Hubot expects all user objects to have a room property that is used in the envelope for the message after it
       # is received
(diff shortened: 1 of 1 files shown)


seratch:
@sinisterstuf Thanks for taking the time to make this PR. LGTM 👍

PR #588, opened 2020-04-01
Fix #586 hubot-slack calling bots.info a suspicious number of times

### LINKED ISSUE #586: hubot-slack calling bots.info a suspicious number of times
### Description

We believe that hubot-slack is calling the `bots.info` API excessively without caching.

We experienced a rate limit event with our Hubot today. After some investigation from Slack support, this is what they found:

> According to our logs, what seems to have happened is that this bot called the `bots.info` API method (https://api.slack.com/methods/bots.info) at a sustained rate of 70 requests/minute for about 40 minutes.

I audited our own code and it doesn't look like we call it directly. I suspect it might be this uncached invocation in hubot-slack, which is outside our application's code: https://github.com/slackapi/hubot-slack/blob/baca8844606c23334fa098d7b48122d7cd4833ef/src/client.coffee#L247-L248

- [x] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements

---

#### Reproducible in:

hubot-slack version: Latest git master

node version: v8.2.0

OS version(s): Debian stretch

#### Steps to reproduce:

Unsure as of yet - because it happens outside our own code we don't know how to cause it to reproduce.

#### Expected result:

hubot-slack works as normal.

#### Actual result:

hubot-slack is rate-limited by Slack.

#### Attachments:

The Slack support request 2659016 may have some additional context for you.

seratch:
###  Summary

This pull request fixes #586 by changing the internals of `SlackClient` a bit.

The latest RTM event payloads from Slack has both `user` and `bot_id` even for events generated by a bot user as below:

```json
{
  "bot_id": "B12345678",
  "suppress_notification": false,
  "type": "message",
  "text": "Badgers? BADGERS? WE DON'T NEED NO STINKIN BADGERS",
  "user": "U12345678",
  "team": "T12345678",
  "bot_profile": {
    "id": "B12345678",
    "deleted": false,
    "name": "hubot",
    "updated": 1568973776,
    "app_id": "A12345678",
    "icons": {
      "image_36": "https://a.slack-edge.com/80588/img/services/hubot_36.png",
      "image_48": "https://a.slack-edge.com/80588/img/services/hubot_48.png",
      "image_72": "https://a.slack-edge.com/80588/img/services/hubot_72.png"
    },
    "team_id": "T12345678"
  },
  "source_team": "T12345678",
  "user_team": "T12345678",
  "channel": "C12345678",
  "event_ts": "1585723843.003300",
  "ts": "1585723843.003300"
}
```

In such cases, `SlackClient`'s `@botUserIdMap` (the internal memory cache of bot user information) is not updated. That's the reason why the cache doesn't work as expected and a large number of `bots.info` calls may occur when receiving bot user events.

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -245,6 +245,7 @@ class SlackClient
     return Promise.resolve(@botUserIdMap[botId]) if @botUserIdMap[botId]?
 
     # Bot user is not in mapping - call bots.info
+    @robot.logger.debug "SlackClient#fetchBotUser() Calling bots.info API for bot_id: #{botId}"
     @web.bots.info(bot: botId).then((r) => r.bot)
 
   ###*
@@ -331,9 +332,13 @@ class SlackClient
     if @eventHandler
       # fetch full representations of the user, bot, and potentially the item_user.
       fetches = {}
-      fetches.user = @fetchUser event.user if event.user
-      fetches.bot = @fetchBotUser event.bot_id if event.bot_id
-      fetches.item_user = @fetchUser event.item_user if event.item_user
+      if event.bot_id
+        fetches.bot = @fetchBotUser event.bot_id
+      else if event.user
+        fetches.user = @fetchUser event.user
+
+      if event.item_user
+        fetches.item_user = @fetchUser event.item_user
 
       # after fetches complete...
       Promise.props(fetches)


src/client.coffee
```diff
@@ -331,9 +332,13 @@ class SlackClient
     if @eventHandler
       # fetch full representations of the user, bot, and potentially the item_user.
       fetches = {}
-      fetches.user = @fetchUser event.user if event.user
-      fetches.bot = @fetchBotUser event.bot_id if event.bot_id
-      fetches.item_user = @fetchUser event.item_user if event.item_user
+      if event.bot_id
+        fetches.bot = @fetchBotUser event.bot_id
+      else if event.user
```
aoberoi, line 337: with this change, when the event has a `bot_id` we will not make a call to Web API method `users.info`, and therefore there won't be a value for `fetched.user` later. 

this seems safe because there's no place in the branch of code where `fetched.bot` is assigned, that the value of `fetched.user` is used. 👍

but i am wondering, how will this reduce the number of calls to `bots.info`?

seratch: @aoberoi 
>but i am wondering, how will this reduce the number of calls to bots.info?

The root cause of the increase of `bots.info` calls is lack of the timing to run the following lines with the latest payloads.

* https://github.com/seratch/hubot-slack/blob/0a063513c4a8c7b53d6481d1b41d35f58622e208/src/client.coffee#L367
* https://github.com/seratch/hubot-slack/blob/0a063513c4a8c7b53d6481d1b41d35f58622e208/src/client.coffee#L372

The combination of the latest payloads and current `SlackClient` implementation keeps the `@botUserIdMap` empty in any case. So, `fetchBotUser` method calls `bots.info` API every time getting events with both `bot_id` and `user`.

aoberoi: ohhh i see! So now that `event.user` is in every bot message, `fetched.user` always has a value, and then `botUserIdMap` is never updated. This change makes it so `fetched.user` will not have a value when both `event.bot` and `event.user` are present. 👍

mistydemeo:
Let me know when you'd like me to test this!

seratch:
@mistydemeo Thanks a lot for your help on this! I just released version 4.7.2. Could you try it out when you have a chance?

mistydemeo:
I've deployed 4.7.2, and it's behaving normally. We deployed at 11:15 PDT; can you please take a look at our API usage and see if it's reduced since then?

seratch:
@mistydemeo Thank you! After your deployment, we're seeing the significant decrease in `bots.info` API calls in data. In the case where you need more help on related matters, I'll be closely working with the support team!

PR #565, opened 2019-05-06
Add compatibility for Coffeescript 2 (and hubot 3)

phucnh:
###  Summary

This PR continues https://github.com/slackapi/hubot-slack/pull/528 works.
Make sure the classes in this code work in Coffeescript 2 (and with hubot 3).

A few of the semantic differences between Coffeescript 1 classes and ES2015 classes has introduced some backwards compatibility issues, as follows:
[Breaking Changes From CoffeeScript 1.x to 2](https://coffeescript.org/#unsupported-get-set)

Known Issue: This PR doesn't make hubot-slack work with hubot 2 in Coffeescript 2. Because hubot 2 still based on Coffeescript 1.

Refs: https://github.com/slackapi/hubot-slack/issues/526

### Requirements (place an `x` in each `[ ]`)

* [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #595, opened 2020-05-05
Fix unexpected newline after super

sinisterstuf:
###  Summary

So I know that the project has specifically pinned coffeescript 1.x in its package.json but I'm trying to use coffeescript 2.x in my project because it has other things I need (namely async/await) and the bot crashes on start, referencing this line from bot.coffee:

    [stdin]:20:10: error: unexpected newline
        super
             ^.

I supposed that this keyword has now become ambiguous without brackets/parentheses but to make sure, I dug up the breaking changes list for the coffeescript 2.0.0 release and found this:

> Due to a syntax clash with `super` with accessors, "bare" `super` (the keyword `super` without parentheses) no longer compiles to a super call forwarding all arguments.
> [...] if you know that the parent function doesn't require arguments, just call `super()`.

Source: https://coffeescript.org/#breaking-changes-super-extends

After adding the brackets the coffeelint tool no longer spits out that error and when running it it "works on my machine".  Tests are still passing and also I can't it imagine it having any adverse effects on users of older versions of coffeescript because the parentheses were optional so far.

### Requirements

* [X] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -17,7 +17,7 @@ class SlackBot extends Adapter
   # @param {Object} options.rtmStart - options for `rtm.start` Web API method
   ###
   constructor: (@robot, @options) ->
-    super
+    super()
     @robot.logger.info "hubot-slack adapter v#{pkg.version}"
     @client = new SlackClient @options, @robot
 


sinisterstuf:
Right, so according to the test results from Travis CI my previous statement about "it works on my machine" isn't good enough because it breaks for Hubot versions < 3.x yay! 😒 🎉

```
TypeError: Cannot read property 'logger' of undefined
```

I am honestly surprised. I suppose the logger it's complaining about is [this one](https://github.com/slackapi/hubot-slack/blob/52c7989149cadf6886a3adab3886ce1c38324cca/src/bot.coffee#L21).

sinisterstuf:
Right, I've just found #565 and it's reverted, so it seems progress to coffeescript 2 is much more complex than I'd hoped.

PR #596, opened 2020-05-05
Add timestamps for enter and leave messages

### LINKED ISSUE #575: Enter/leave messages don't have timestamps
### Description

Marking this as question because I'm not sure it's a bug. I've noticed that the enter and leave events are missing a timestamp (I will talk about only enter events from here on because leave events behave the same).

Most things in the eventHandler return a timestamp from `event.ts` or `event.event_ts`. Enter events do not: ([source](https://github.com/slackapi/hubot-slack/blob/master/src/bot.coffee#L262-L266))

```coffeescript
    else if event.type is "member_joined_channel"
      # this event type always has a channel
      user.room = channel
      @robot.logger.debug "Received enter message for user: #{user.id}, joining: #{channel}"
      @receive new EnterMessage user
```

The only thing it passes to `@receive` is user, and you can see as much if you console.log what you get when using `robot.enter` as a script developer; you have user data and that's it.

My use case is: I'd like the robot to apply emoji reactions to the message for the enter event in Slack but that action [requires a timestamp and a channel](https://api.slack.com/methods/reactions.add#arguments). I can get the channel from room, kinda, but I don't have the message timestamp.

To be honest it's not clear to me whether that RTM event even contains a timestamp or not because the documentation for [member_joined_channel](https://api.slack.com/events/member_joined_channel) does not include a timestamp in the example even though the example in [the 2017 blog post introducing this event](https://api.slack.com/changelog/2017-05-rethinking-channel-entrance-and-exit-events-and-messages) _does_ include one. What's for sure is that the [message/channel_join](https://api.slack.com/events/message/channel_join) _did_ include a timestamp and probably its frequency would be sufficient for my use case but we moved away from using that in #513.

If the RTM event member_joined_channel does include a timestamp, this is perfect and I will create a pull request to try to propogate that timestamp to Hubot. How can I find that out?

If it does _not_ have one, then it is probably intentional and in that case, how can I work around the now-missing message/channel_join which has the timestamp I need?

Sorry for the wall of text 😅

- [x] question

### Requirements

sinisterstuf:
###  Summary

Enter and leave messages don't have timestamps, unlike other events like text messages.  Hubot doesn't have a place for them e.g. for enter messages in the EnterMessage class, but Slack supports them and other actions regarding enter/leave events in Slack expect you to use a timestamp, so you need to be able to get it from somewhere.

This change works around the fact that they're not present in Hubot's classes by inserting a `ts` field after the message object has been created.  It includes tests to confirm that Slack's event timestamp is successfully passed on to Hubot.

Resolves #575

### Requirements

* [X] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/hubot-slack/blob/master/.github/contributing.md) and have done my best effort to follow them.

### DIFF
--- a/src/bot.coffee
+++ b/src/bot.coffee
@@ -272,13 +272,17 @@ class SlackBot extends Adapter
       # this event type always has a channel
       user.room = channel
       @robot.logger.debug "Received enter message for user: #{user.id}, joining: #{channel}"
-      @receive new EnterMessage user
+      msg = new EnterMessage user
+      msg.ts = event.ts
+      @receive msg
 
     else if event.type is "member_left_channel"
       # this event type always has a channel
       user.room = channel
       @robot.logger.debug "Received leave message for user: #{user.id}, joining: #{channel}"
-      @receive new LeaveMessage user
+      msg = new LeaveMessage user
+      msg.ts = event.ts
+      @receive msg
 
     else if event.type is "reaction_added" or event.type is "reaction_removed"
 
--- a/test/bot.coffee
+++ b/test/bot.coffee
@@ -223,13 +223,23 @@ describe 'Handling incoming messages', ->
     return
 
   it 'Should handle member_joined_channel events as envisioned', ->
-    @slackbot.eventHandler {type: 'member_joined_channel', user: @stubs.user, channel: @stubs.channel.id}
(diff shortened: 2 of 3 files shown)


sinisterstuf:
Thanks! 😁

